PR #13, opened 2022-08-19
Refactor nxsunit

### LINKED ISSUE #5: The celsius/kelvin conversion in nxsunit is not correct.
Celsius and kelvin temperature units are currently considered equivalent, with no offset between them.

A related implementation from another project is attached, with support for celsius, kelvin and Fahrenheit.

[nxsunit.txt](https://github.com/SasView/sasview/files/4714202/nxsunit.txt)

krzywon:
This overhauls the nxsunit module to allow scaling with offsets, and scaling of multi-dimensional and complex units, e.g. SESANS A^-2*cm^-1 will treat the A^-2 and cm^-1 as separate components, each independently scalable. Unit tests were moved from within the module to the global unit test suite.

Fixes #5
Fixes(?) sasview/sasview#2035 - @pkienzle what do you think?

### DIFF
--- a/sasdata/data_util/nxsunit.py
+++ b/sasdata/data_util/nxsunit.py
@@ -43,12 +43,13 @@ allow the application to set the dimension for the unit rather than
 inferring the dimension from an example unit.
 """
 
-from __future__ import division
 import math
 import re
-
+from typing import Dict, Union, TypeVar, Tuple, Sequence, Optional, List
 
 __all__ = ['Converter']
+T = TypeVar('T')
+ConversionType = Union[float, Tuple[float, float]]
 DIMENSIONS = {}  # type: Dict[str, Dict[str, ConversionType]]
 AMBIGUITIES = {}  # type: Dict[str, str]
 PREFIX = dict(peta=1e15, tera=1e12, giga=1e9, mega=1e6, kilo=1e3, deci=1e-1, centi=1e-2, milli=1e-3, mili=1e-3,
@@ -59,8 +60,7 @@ SHORT_PREFIX = dict(P=1e15, T=1e12, G=1e9, M=1e6, k=1e3, d=1e-1, c=1e-2, m=1e-3,
 # Limited form of units for returning objects of a specific type.
 # Maybe want to do full units handling with e.g., pyre's
 # unit class. For now lets keep it simple.  Note that
-def _build_metric_units(unit, abbr):
-    # type: (str, str) -> dict[str, float]
+def _build_metric_units(unit: str, abbr: str) -> Dict[str, float]:
     """
     Construct standard SI names for the given unit.
     Builds e.g.,
@@ -74,7 +74,7 @@ def _build_metric_units(unit, abbr):
 
     Returns a dictionary of names and scales.
     """
-    map = {}
+    units = {}
     for name in [unit, unit.capitalize(), unit.lower(), abbr]:
         for items in [PREFIX, SHORT_PREFIX]:
             names = {}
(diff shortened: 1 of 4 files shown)


butlerpd:
Still waiting for review by @pkienzle and @lucas-wilkins (though the latter will be after Jan 1 I guess?). @wpotrzebowski may also take a look.

sasdata/data_util/nxsunit.py
```diff
@@ -43,173 +43,405 @@
...
+    unit = re.sub(r'sec(ond)?(s)?', 's', unit, flags=re.IGNORECASE)
+    # Catch kelvin, KELVIN, and any capitalization in between
+    # Replace with 'K'
+    unit = re.sub(r'kel(vin)?(s)?', 'K', unit, flags=re.IGNORECASE)
+    # Catch celcius, CELCIUS, and any capitalization in between
+    # Replace with '℃'
+    unit = re.sub(r'cel(cius)?', '℃', unit)
+    # Catch hertz, HERTZ, hz, HZ, and any capitalization in between
+    # Replace with 'Hz'
+    unit = re.sub(r'h(ert)?z', 'Hz', unit)
```
wpotrzebowski, line 294: Is Hertz/Hz also allowed?

sasdata/data_util/nxsunit.py
```diff
@@ -43,173 +43,405 @@
...
+                    for k, v in DIMENSIONS.items():
+                        if unit in v:
+                            self.dimension.append(k)
+                            break
+                    else:
+                        self.dimension.append('dimensionless')
+
+        # Find the scale for the given units - default to dimensionless
+        self.scalemap = [DIMENSIONS.get(dimension, DIMENSIONS['dimensionless']) for dimension in self.dimension]
+        base = self._get_scale_for_units(self._units)
```
wpotrzebowski, line 397: Maybe better to return scalebase and scaleoffset as two separate variables?

wpotrzebowski (commented):
I am not that familiar with this code base but code looks good. Minor stylistic remark There are few places with for loops with single letter variables. I haven't tested functionality yet.

sasdata/data_util/nxsunit.py
lucas-wilkins: think they need the ignorecase flags set

lucas-wilkins (approved):
Looks OK

sasdata/data_util/nxsunit.py
```diff
@@ -43,173 +43,405 @@
...
     """
     Construct names for the given units.  Builds singular and plural form.
     """
-    map = {}
-    map.update([(name,scale) for name,scale in kw.items()])
-    map.update([(name+'s',scale) for name,scale in kw.items()])
-    return map
+    units = {}
+    units.update([(name, scale) for name, scale in kw.items()])
+    units.update([(name + 's', scale) for name, scale in kw.items()])
```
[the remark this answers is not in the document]
krzywon: I agree, but this class works mostly in SI units. Here is a list of edge cases we might come up against where this fails, based on my understanding of [the NIST guide to SI](https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-9-rules-and-style-conventions-spelling-unit-names):

- Anything Imperial that isn't a temperature
  - Potential solution: add imperial lengths and weights
  - What I am willing to do for this PR: I will continue to omit this. The scattering community doesn't work in inch(es)/feet/pounds/etc.
- `microohm` and `kiloohm` will be generated by the class, but `microhm` and `kilohm` are the correct forms.
  - Potential solution: explicitly add these as resistance units
  - What I am willing to do for this PR: I could implement this pretty quickly if there is a call for it.
- units in long-form, e.g. `meters per second squared` -or- `meters per square second`
  - Potential solution 1: convert `<square|cubic> <unit>` or `<unit> <squared|cubed>` to `unit^<2|3>`
  - Potential solution 2: convert `per` to `/`
  - What I am willing to do for this PR: I ignored this because long-form is typically reserved for textual content, not data units, but implementation would be fairly quick. I'd also need to add some reasonable unit tests.

butlerpd:
is this ready to merge?

PR #15, opened 2022-09-06
14: Rework readme

### LINKED ISSUE #14: Incorrect virtual environment instructions in readme
From @celinedurniak:

> I think that there are a few issues with the installation instructions in https://github.com/SasView/sasdata/blob/master/README.md
> 1. The virtual environment should be activated before installing the modules via pip.
> 2. The command to activate the environment is OS-dependent. For MacOS, it would be source sasdata/bin/activate

krzywon:
This updates the readme to accommodate the changes suggested by @celinedurniak.

Fixes #14

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,31 +1,28 @@
 # Sasdata
 
-A package for loading and handling reduced small angle scattering data.
+A package for importing and exporting reduced small angle scattering data.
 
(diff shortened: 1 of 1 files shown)


celinedurniak:
The new instructions to use sasdata with a virtual environment or conda are good to be merged.

celinedurniak (approved):

PR #28, opened 2022-09-28
Generalize the Data object

### LINKED ISSUE #17: Generalize 1D and 2D data
As noted in sasview/sasview#841, the 1D and 2D data objects could be abstracted out into a single class. This would allow more generic data formats to be added as needed.

In this package, the following is required:

- [ ] Combine the plottable_1d and plottable_2d classes into a single class (called plottable?)
- [ ] Combine the Data1D and Data2D classes into a single class (called Data?)
- [ ] Update all base file loader class
- [ ] Update all specific reader classes
- [ ] Update file conversion routines
- [ ] Update manipulations (currently only works for 2D data and there are checks for this)
- [ ] Update unit tests to match new functionality

### LINKED ISSUE #19: Add mask to 1D Data
As noted in sasview/sasview#1113, the plottable_1d class has no mask property. This should be added, but also would be handled with #17.

krzywon:
@lucas-wilkins - this is a start to what I am imagining for how the data object should be refactored. I need this to finish the separation of sasview/sasview#1802 related to unit conversions but want a third set of eyes (@pkienzle already took a quick look and gave his input) to take a look at the direction I am taking.

Once finished, this will fix #17, and fix #19.


lucas-wilkins:
Cool, I have a workshop tomorrow and Friday, but I might find some time to look somewhere.

lucas-wilkins:
I think its going in the right direction. Would be good to have a chat about it soon.

butlerpd:
Working on cleanup in preparation for Contributor Camp XII:

@krzywon I'm assuming this kind of change should probably go  into 6.0? in which case we may need to prioritize it as such? or is that not the case?

krzywon:
> I'm assuming this kind of change should probably go into 6.0? in which case we may need to prioritize it as such? or is that not the case?

Ideally, yes, but realistically, probably won't make it in.

krzywon:
Superseded by #79

PR #36, opened 2023-03-20
Enable the sector slicing to allow both sides independantly

butlerpd:
NOTE: the commit message was left over from a previous commit. This commit effectively allows `manipulations.py` to respond to a new parameter (self.fold) to decide whether to "fold" the two sectors into one or return negative qs for the lower sector so as to keep the two sectors separate.

This should be merged prior to SasView/sasview#1919 which will try to set the fold parameter.

### DIFF
--- a/sasdata/data_util/manipulations.py
+++ b/sasdata/data_util/manipulations.py
@@ -65,7 +65,8 @@ def get_q_compo(dx: float, dy: float, detector_distance: float, wavelength: floa
 
 def flip_phi(phi: float) -> float:
     """
-    Correct phi to within the 0 <= to <= 2pi range
+    Force phi to be within the 0 <= to <= 2pi range by adding or subtracting
+    2pi as necessary
 
     :return: phi in >=0 and <=2Pi
     """
@@ -282,8 +283,14 @@ class Binning:
     def __init__(self, min_value, max_value, n_bins, base=None):
         """
         if base is None: Linear binning
+        Todo: should base be a boolean? True = log? or do we want to allow for
+              other types of binning in the future, in which case we should
+              define base="linear" or "log" and explicitly set
+              if base = None then base="log"
         """
-        self.min = min_value if min_value > 0 else 0.0001
+        # min and max values can be negative.
+        # todo: do we want to check for min < max?
+        self.min = min_value
         self.max = max_value
         self.n_bins = n_bins
         self.base = base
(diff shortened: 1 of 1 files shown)


sasdata/data_util/manipulations.py
```diff
@@ -282,8 +283,14 @@ class Binning:
     def __init__(self, min_value, max_value, n_bins, base=None):
         """
         if base is None: Linear binning
+        Todo: should base be a boolean? True = log? or do we want to allow for
```
krzywon, line 286: An Enum would probably be a better option to pre-define the binning types available. This would offer extensibility as well. Work for this is likely outside the scope of this PR.

sasdata/data_util/manipulations.py
```diff
@@ -282,8 +283,14 @@ class Binning:
...
+              other types of binning in the future, in which case we should
+              define base="linear" or "log" and explicitly set
+              if base = None then base="log"
         """
-        self.min = min_value if min_value > 0 else 0.0001
+        # min and max values can be negative.
+        # todo: do we want to check for min < max?
```
krzywon, line 292: Annular averaging gives the exact same result if `q_min > q_max`. There is no way to do this for sector cuts and slabs through the GUI. A unit test to show this is true for all slicer models could be useful...

krzywon (approved):
The code looks fine but could benefit from some unit tests. I did not test functionality.

sasdata/data_util/manipulations.py
butlerpd: I am a bit confused about what point is bein made here. For a true annular (phi average) it makes no sense to have q < 0 (max or min) but otherwise does. I don't think flip gets called unless there is a limited phi range in which case allowing q < makes sense.  However I have noted that all slicer interactors seem to allow max < min and have been trying to figure out if that is a feature that is brilliant or a bug that needs squashing.  Is it the last that you are concerned about here? or something else?

krzywon: I'm saying the check you are suggesting, to see if `q_max > q_min`, isn't needed. Mathematically, there is no difference in the resulting calculation regardless of which is larger. I calculated the annular averaging slicer twice, where the min and max q values are swapped for the second calculation, and the results are exactly the same. I don't know if this is true for other slicers, but that's where my unit test suggestion would come into play. Run the calculations for every slicer twice as I did for the annular averaging then compare the results.

sasdata/data_util/manipulations.py
butlerpd: Have now updated that comment

sasdata/data_util/manipulations.py
butlerpd: Ok ... I see the confusion. Yes I know it works both ways my question was really "should it"? I have rewritten my comment to accurately reflect the question.

butlerpd:
@krzywon, I believe we said at one of the calls that as soon as I updated the first todo we could merge this. Can we resolve the conversations now?

PR #37, opened 2023-04-11
31: Load data from URIs

### LINKED ISSUE #31: load data from url
It might be convenient to load data from a url instead of a file.

Something like:
```python
def open_or_fetch(uri):
  if '://' in uri:
    req = requests.get(uri)
    req.raise_for_status()
    #print(req.content)
    fd = BytesIO(req.content)
    fd.name = uri
    return fd
  else:
    return open(uri, 'rb')
...
class FileReader:
    ...
    def read(self, filepath):
        ...
        if '://' in filepath or os.path.isfile(filepath):
            ...
            with open_or_fetch(filepath) as self.f_open:
                self.get_file_contents()
```

By monkey-patching I was able to get to load and plot using the following on colab::
```python
from sasdata.dataloader import loader as loader_mod
from sasmodels.data import plot_data
from importlib import reload
reload(loader_mod) # need reload the module whenever FileReader changes.

root = 'https://raw.githubusercontent.com/SasView/sasview/main/src/sas/example_data/'
#uri = root + '1d_data/AOT_Microemulsion-Core_Contrast.xml'
#uri = root + '1d_data/FK403_0006_Nika.hdf'
#uri = root + '1d_data/ testdata_line1.txt'
uri = root + '1d_data/apoferritin.txt'
#uri = root + '1d_data/AUSANS_run3_2_no_buffer.ABS'
loader = loader_mod.Loader()
data = loader.load(uri)[0]
print(data)
data.mask = np.zeros_like(data.x, dtype='bool')
plot_data(data)
```

Here's the full patch cell I used to get it working:
```python
#%pip install sasmodels sasdata matplotlib bumps periodictable
import os
from io import BytesIO
import requests
from sasdata.dataloader.filereader import FileReader, DEPRECATION_MESSAGE
from sasdata.data_util.loader_exceptions import NoKnownLoaderException, FileContentsException,\
    DataReaderException

def open_or_fetch(uri):
  if '://' in uri:
    req = requests.get(uri)
    req.raise_for_status()
    #print(req.content)
    fd = BytesIO(req.content)
    fd.name = uri
    return fd
  else:
    return open(uri, 'rb')


def read(self, filepath: str): #-> List[Union[Data1D, Data2D]]:
        """
        Basic file reader

        :param filepath: The full or relative path to a file to be loaded
        """
        self.filepath = filepath
        if '://' in filepath or os.path.isfile(filepath):
            basename, extension = os.path.splitext(os.path.basename(filepath))
            self.extension = extension.lower()
            # If the file type is not allowed, return nothing
            if self.extension in self.ext or self.allow_all:
                # Try to load the file, but raise an error if unable to.
                try:
                    print(f"++++++++++ trying {self.extension}")
                    with open_or_fetch(filepath) as self.f_open:
                        self.get_file_contents()
                except DataReaderException as e:
                    print(f"********** {self.extension} {str(e)}")
                    self.handle_error_message(str(e))
                except FileContentsException as e:
                    print(f"********** {self.extension} {str(e)}")
                    raise
                except OSError as e:
                    print(f"********** {self.extension} {str(e)}")
                    # If the file cannot be opened
                    msg = "Unable to open file: {}\n".format(filepath)
                    msg += str(e)
                    self.handle_error_message(msg)
                except Exception as e:
                    print(f"********** {self.extension} {str(e)}")
                    self.handle_error_message(str(e))
                finally:
                    print(f"Finally {self.extension}")
                    if any(filepath.lower().endswith(ext) for ext in
                           self.deprecated_extensions):
                        self.handle_error_message(DEPRECATION_MESSAGE)
                    if len(self.output) > 0:
                        # Sort the data that's been loaded
                        self.convert_data_units()
                        self.sort_data()
        else:
            msg = "Unable to find file at: {}\n".format(filepath)
            msg += "Please check your file path and try again."
            self.handle_error_message(msg)

        # Return a list of parsed entries that data_loader can manage
        final_data = self.output
        self.reset_state()
        return final_data

FileReader.read = read
```

HDF, xml and abs didn't work; not sure why.

krzywon:
This PR adds a few non-breaking enhancements to the data loader system to aid in scripting and web-facing capabilities.

- Fixes #31: The Loader class now recognizes URI strings, and, using the python built-in module `urllib.request`, opens the file at from the web address.
- Refs sasview/sasview#1845: Lists of files can now be sent to the loader instead of just a single file and all loaded Data objects will be returned as a list. **Note**: This functionality is not available to individual readers.
- The loader is directly callable to allow transient instances of the loader object to aid in scripting.
- File handles are passed to the readers instead of file names so data files (especially remote files) are only opened a single time, regardless of the number of loaders attempting to read the file, while maintaining backward compatibility when passing a file path directly to a reader.
- Unit tests were added
  - Comparing file contents loaded from the local repository to files loaded from their GitHub URIs for the following formats: ABS, Anton-Parr SAXSess PDH, canSAS XML, NXcanSAS, 2D NIST DAT, and SESANS.
  - Comparing file contents loaded individually as compared to files loaded as a list.
  - Ensuring Readers and Loader behavior matched for individual files
- The ReadMe was updated to match the new functionality.


butlerpd:
@pkienzle said he will look at this

sasdata/dataloader/filereader.py
```diff
@@ -74,62 +76,71 @@ def __init__(self):
...
-                    self.handle_error_message(str(e))
-                finally:
-                    if any(filepath.lower().endswith(ext) for ext in
-                           self.deprecated_extensions):
-                        self.handle_error_message(DEPRECATION_MESSAGE)
-                    if len(self.output) > 0:
-                        # Sort the data that's been loaded
-                        self.convert_data_units()
-                        self.sort_data()
+        if isinstance(file_handler, CustomFileOpen):
```
pkienzle, line 97: If you need `isinstance` then your object structure is broken.

Also, copying private handles out of the `CustomFileOpen` is bad practice. The handle lifetime is controlled by `CustomFileOpen`, and may be closed and deleted before `FileReader` is done using it.

sasdata/dataloader/filereader.py
```diff
@@ -74,62 +76,71 @@ def __init__(self):
...
+            self.f_open = file_handler.fd
+            self.hdf_open = file_handler.h5_fd
+            self.string_open = file_handler.string_fd
         else:
-            msg = "Unable to find file at: {}\n".format(filepath)
-            msg += "Please check your file path and try again."
+            # If not called by a context manager, create one locally that calls read() again. Return early
+            return self._read_with_local_context_manager(filepath)
```
pkienzle, line 103: This gets confusing to debug.

I suggest either creating a new reader class than accepts a byte stream and replace the existing with a shim to call the new reader to preserve the interface, or perhaps a new method that accepts a byte stream, with the existing read method opening the byte stream and returning a call to the new method.

sasdata/dataloader/filereader.py
```diff
@@ -138,21 +149,23 @@ def nextline(self) -> str:
         Returns the next line in the file as a string.
         """
         #return self.f_open.readline()
-        return decode(self.f_open.readline())
+        return decode(self.string_open.readline())
```
pkienzle, line 152: Isn't `string_open` already decoded?

sasdata/dataloader/filereader.py
```diff
@@ -138,21 +149,23 @@ def nextline(self) -> str:
...
             yield decode(line)
 
     def readall(self) -> str:
         """
         Returns the entire file as a string.
         """
-        return decode(self.f_open.read())
+        handle = self.string_open if self.string_open else self.f_open
+        handle.seek(0)
+        return decode(handle.read())
```
pkienzle, line 168: If `read` is called directly instead of `readall` then `seek` will not be triggered and the file may be in the wrong position. So go back to my original position: have the registry reset the seek position between reader attempts.

sasdata/dataloader/loader.py
```diff
@@ -363,14 +364,26 @@ def associate_file_reader(self, ext: str, loader) -> bool:
         """
         return self.__registry.associate_file_reader(ext, loader)
 
-    def load(self, file: str, format: Optional[str] = None) -> Union[list, Exception]:
+    def load(self, file_path_list: Union[List[str], str],
+             format: Optional[Union[List[str], str]] = None
+             ) -> Union[List[Union[Data1D, Data2D]], Exception]:
```
pkienzle, line 369: You are always returning a list. Exceptions are being eaten, which is bad. If the registry load is set up to return rather than raise exceptions, then the return signature would be `List[Union[Data1D, Data2D, Exception]]`.

sasdata/dataloader/loader.py
```diff
@@ -363,14 +364,26 @@ def associate_file_reader(self, ext: str, loader) -> bool:
...
         """
-        Load a file
-        :param file: file name (path)
+        Load a file or series of files
+        :param file_path_list: String representations of any number of file paths. This can either be a list or a string
         :param format: specified format to use (optional)
         :return: DataInfo object
         """
-        return self.__registry.load(file, format)
+        if isinstance(file_path_list, str):
```
pkienzle, line 376: What if you are called with a `pathlib.Path` object? Though that might have other issues.

I've been moving away from interfaces that accept either a singleton or a list toward separate functions for singletons and lists. However, in the list case you may want to accept a single format or a sequence of formations of the same length.

You can loop over `zip(file_path_list, format_list)` rather than using an index into `format_list`.

sasdata/dataloader/loader.py
```diff
@@ -400,3 +413,11 @@ def get_wildcards(self):
         Return the list of wildcards
         """
         return self.__registry.wildcards
+
+    def __call__(self, file_path_list: List[str]) -> List[Union[Data1D, Data2D]]:
```
pkienzle, line 417: Return signature should match the return signature of `self.load`.

sasdata/dataloader/readers/anton_paar_saxs_reader.py
```diff
@@ -62,7 +62,7 @@ def get_file_contents(self):
 
         # Reinitialize the class when loading a new data file to reset all class variables
         self.reset_state()
-        buff = self.readall()
+        buff = self.readall().strip()
```
pkienzle, line 65: This only strips white space from the start and end of the text, not for each line. Are you sure you don't want to handle blank lines/comment lines as part of the normal line-by-line processing of raw data?

Using the `raw_data` attribute to hand data to `read_data` is a bit odd. With a name like `read_data` I would expect it to read the data, but at that point the data has already been read. But that's outside the scope of this PR.

sasdata/dataloader/readers/cansas_reader_HDF5.py
```diff
@@ -66,41 +66,18 @@ def get_file_contents(self):
...
-
-                for data_set in self.output:
-                    if isinstance(data_set, Data1D):
-                        if data_set.x.size < 5:
-                            exception = FileContentsException(
-                                "Fewer than 5 data points found.")
-                            data_set.errors.append(exception)
+        try:
+            # Read in all child elements of top level SASroot
+            self.read_children(self.hdf_open, [])
```
pkienzle, line 71: Any reason not to open an HDF file here using the byte stream, then close it when you are done reading children?

sasdata/dataloader/readers/xml_reader.py
```diff
@@ -68,40 +80,55 @@ def reader(self):
...
+                self._set_xml_string(xml)
         except etree.XMLSyntaxError as xml_error:
             logger.info(xml_error)
             raise xml_error
-        except Exception:
+        except Exception as exc:
             self.xml = None
             self.xmldoc = None
             self.xmlroot = None
+            raise exc
```
pkienzle, line 103: A bare `raise` without an exception re-raises the current exception.

sasdata/dataloader/loader.py
```diff
@@ -363,14 +327,16 @@ def associate_file_reader(self, ext: str, loader) -> bool:
         """
         return self.__registry.associate_file_reader(ext, loader)
 
-    def load(self, file: str, format: Optional[str] = None) -> Union[list, Exception]:
+    def load(self, file_path_list: Union[List[Union[str, Path]], str, Path],
+             format: Optional[Union[List[str], str]] = None
+             ) -> Union[List[Union[Data1D, Data2D]]]:
```
pkienzle, line 332: Extra `Union[...]` wrapper in the return type.

sasdata/dataloader/readers/ascii_reader.py
```diff
@@ -46,7 +47,7 @@ def get_file_contents(self):
         """
 
         buff = self.readall()
-        filepath = self.f_open.name
+        filepath = self.filepath.name
```
pkienzle, line 50: Tempting to call this filename instead of filepath.

sasdata/dataloader/readers/sesans_reader.py
```diff
@@ -77,13 +77,13 @@ def get_file_contents(self):
...
 
-        data = np.loadtxt(self.f_open)
+        data = np.loadtxt(self.nextlines())
 
         if data.shape[1] != len(headers):
             raise FileContentsException(f"File has {len(headers)} headers, but {data.shape[1]} columns")
 
         if not data.size:
-            raise FileContentsException("{} is empty".format(self.filepath))
+            raise FileContentsException("{} is empty".format(self.filepath.name))
```
pkienzle, line 86: Since you are modifying the line anyway, might as well use and f-string.

README.md
```diff
@@ -27,29 +27,52 @@ or [anaconda](https://www.anaconda.com/)::
...
+  - `.h5`, `.hdf`, `.hdf5`, `.nxs`: [NXcanSAS](https://manual.nexusformat.org/classes/applications/NXcanSAS.html) format
+  - `.txt`: Multi-column ascii format
+  - `.csv`: Comma delimited text format
+  - `.ses`, `.sesans`: [Multi-column SESANS](https://www.sasview.org/docs/user/qtgui/MainWindow/data_formats_help.html#d-sesans-format) data
+  - `.dat`: [2D NIST](https://github.com/sansigormacros/ncnrsansigormacros/wiki/NCNROutput2D_QxQy) format
+  - `.abs`, `.cor`: 1D NIST format for SAS and USAS
+  - `.pdh`: Anton Paar reduced SAXS format
+
+The `save()` method accepts 3 arguments; the file path to save the file as, a `Data1D` or `Data2D` object, and, optionally, 
+a file extension. If an extension is passed to `save`, any file extension in the file path will be superseded.
```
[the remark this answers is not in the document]
krzywon: `save('mydata.xml', format='.csv')` outputs the file `mydata.xml.csv` in the CSV format. The format **always** supersedes the filename.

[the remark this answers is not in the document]
pkienzle: My questions were rhetorical, inviting you to update the doc string to describe the behaviour  8-)

sasdata/dataloader/filereader.py
```diff
@@ -74,47 +76,61 @@ def __init__(self):
         self.current_datainfo = None
         # File path sent to reader
         self.filepath = None
+        # Starting file position to begin reading data from
+        self.f_pos = 0
+        # File extension sent to reader
+        self.extension = None
```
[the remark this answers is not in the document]
krzywon: Good catch. The scope of this PR has grown too much at this point, so I will submit an issue for this and fix it somewhere else.

sasdata/dataloader/filereader.py
```diff
@@ -74,47 +76,61 @@ def __init__(self):
...
+        basename, extension = self.filepath.stem, self.filepath.suffix
+        self.extension = extension.lower()
+        if self.extension in self.ext or self.allow_all:
+            try:
+                self.get_file_contents()
+            except Exception as e:
+                raise
+            finally:
+                if any(self.filepath.name.lower().endswith(ext) for ext in
+                       self.deprecated_extensions):
```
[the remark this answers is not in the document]
pkienzle: Should also list filename in warning in case there is a list of files being loaded.

sasdata/dataloader/filereader.py
```diff
@@ -74,47 +76,61 @@ def __init__(self):
...
+                    self.handle_error_message(DEPRECATION_MESSAGE)
+                if len(self.output) > 0:
+                    # Sort the data that's been loaded
+                    self.convert_data_units()
+                    self.sort_data()
```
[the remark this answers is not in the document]
krzywon: I added developer documentation explaining the specific use case this try/finally is handling, but agree with your assessment. I've added an issue to come back to the edge case where a file could be loaded multiple times.

pkienzle (approved):
Tested with pytest and `python -m sasmodels.compare sphere -data=https://github.com/SasView/sasdata/raw/master/test/sasdataloader/data/valid_cansas_xml.xml`

I still think the CustomFileOpen is unnecessarily complicated but it works as is, and it is just as much effort to change in a future PR as it is to change in this PR.

PR #41, opened 2023-08-02
Fixing Issue #40 - SectorPhi class throws errors (bin index out of range)

### LINKED ISSUE #40: SectorPhi class throws errors (bin index out of range) when phi range includes 0 (or 2Pi)
**Describe the bug**
The title describes how the bug appears to the user. The source of the problem is the method `Binning.get_bin_index()` in `sasdata/data_util/manipulations.py`.
If used for angular coordinates, it is possible for the bin index it returns to be greater than the total number of bins, as the subtraction operation doesn't account for the cyclic nature of angles. An error is later thrown when this bad bin index is used, causing the "Wedge Averaging in Phi" slicer to break (this is the only slicer to use the SectorPhi class).

**To Reproduce**
1. In SasView, Compute/Plot any 2D data from `example_data/2d_data`, using any model.
2. Right click the plot and select 'Wedge Averaging in Phi'
3. Drag and drop the central line so that the wedge region passes over the negative x-axis.
4. See that the slicer cannot be placed here, as it continues to move with the mouse cursor. Also note the error shown in the terminal.

**Expected behavior**
The slicer should cease to move when the mouse button is released, and there should be no error thrown.

ehewins:
This commit makes the necessary changes to fix the issues with "Wedge Averaging in Phi" using the Wedge Slicer, where certain phi values would cause a 'bin index out of range' error.
The method `Binning.get_bin_index()` from `manipulations.py` now checks if the index it's about to return is bad, and corrects it if so (using an improved version of the flip_phi function).

Fixes #40 

The steps needed to produce the original error have been followed, and the error is now fixed. "Wedge Averaging in Q" was also tested to see if these changes introduce errors in other slicers using the new `get_bin_index()` method, but no problems were found.

### DIFF
--- a/sasdata/data_util/manipulations.py
+++ b/sasdata/data_util/manipulations.py
@@ -63,18 +63,19 @@ def get_q_compo(dx: float, dy: float, detector_distance: float, wavelength: floa
     return out
 
 
-def normalize_angle(phi: float, closed_at_end: bool = False) -> float:
+def normalize_angle(phi: float) -> float:
     """
-    Force phi to be within the half-open interval 0 <= phi < 2pi, unless
-    closed_at_end is set to True, in which case the interval is 0 < phi <= 2pi.
-
-    :param phi: angle to return in a normalized form (0 <= phi < 2pi)
-    :param closed_at_end: if True, changes the interval to 0 < phi <= 2pi
+    Returns the supplied angle in the range 0 <= phi <= 2pi.
+    When phi is a non-zero integer multiple of 2pi, it is more appropriate to
+    return 2pi than zero. This allows rings to be defined with start points
+    at 0 and end points at 2pi.
     """
-    phi_out = phi % (2 * np.pi)
-    if closed_at_end and phi_out == 0:
-        phi_out = 2 * np.pi
-
+    quotient, remainder = divmod(phi, 2*np.pi)
+    # If phi is a non-zero integer multiple of 2pi:
+    if remainder == 0 and quotient > 0:
+        phi_out = 2*np.pi
+    else:
+        phi_out = remainder
     return phi_out
 
 
@@ -302,16 +303,18 @@ class Binning:
         bin = floor(N * (log(x) - log(min)) / (log(max) - log(min)))
         """
         if self.base:
-            temp_x = math.log(value, self.base) - math.log(self.min, self.base)
-            temp_y = math.log(self.max, self.base) - math.log(self.min, self.base)
+            numerator = math.log(value, self.base) - math.log(self.min, self.base)
+            denominator = math.log(self.max, self.base) - math.log(self.min, self.base)
         else:
-            temp_x = value - self.min
-            temp_y = self.max - self.min
-        # Fixing an issue where certain angular coordinate values give bad bin values
-        if np.fabs(temp_x) > np.fabs(temp_y):
-            temp_x = normalize_angle(temp_x)
+            numerator = value - self.min
+            denominator = self.max - self.min
+        # This if statement can only trigger if value > self.max, which should
+        # only ever happen when binning an angular parameter.
+        # Calling normalize_angle() ensures the bin index returned <= n_bins.
+        if np.fabs(numerator) > np.fabs(denominator):
+            numerator = normalize_angle(numerator)
         # Bin index calulation
-        return int(math.floor(self.n_bins * temp_x / temp_y))
+        return int(math.floor(self.n_bins * numerator / denominator))
 
 
 ################################################################################
@@ -873,13 +876,13 @@ class _Sector:
         x_err = np.zeros(self.nbins)
         y_counts = np.zeros(self.nbins)  # Cycle counts (for the mean)
 
-        # Get the min into region: 0 <= phi < 2Pi and max into region: 0 < phi <= 2Pi
+        # Get the min and max into the region: 0 <= phi < 2Pi
         phi_min = normalize_angle(self.phi_min)
-        phi_max = normalize_angle(self.phi_max, closed_at_end=True)
+        phi_max = normalize_angle(self.phi_max)
         # Now calculate the angles for the opposite side sector, here referred
         # to as "minor wing," and ensure these too are within 0 to 2pi
         phi_min_minor = normalize_angle(phi_min - math.pi)
-        phi_max_minor = normalize_angle(phi_max - math.pi, closed_at_end=True)
+        phi_max_minor = normalize_angle(phi_max - math.pi)
 
         #  set up the bins by creating a binning object
         if run.lower() == 'phi':
@@ -1196,9 +1199,9 @@ class Sectorcut:
         # get phi from data
         phi_data = np.arctan2(qy_data, qx_data)
 
-        # Get the min into region -pi <= phi < pi and max into region -pi < phi <= pi
+        # Get the min and max into the region: -pi <= phi < Pi
         phi_min_major = normalize_angle(self.phi_min + Pi) - Pi
-        phi_max_major = normalize_angle(self.phi_max + Pi, closed_at_end=True) - Pi
+        phi_max_major = normalize_angle(self.phi_max + Pi) - Pi
         # check for major sector
         if phi_min_major > phi_max_major:
             out_major = (phi_min_major <= phi_data) + \
@@ -1208,9 +1211,9 @@ class Sectorcut:
                 phi_max_major > phi_data)
 
         # minor sector
-        # Get the min into region -pi <= phi < pi and max into region -pi < phi <= pi
+        # Get the min and max into the region: -pi <= phi < Pi
         phi_min_minor = normalize_angle(self.phi_min) - Pi
-        phi_max_minor = normalize_angle(self.phi_max, closed_at_end=True) - Pi
+        phi_max_minor = normalize_angle(self.phi_max) - Pi
 
         # check for minor sector
         if phi_min_minor > phi_max_minor:
--- a/sasdata/data_util/manipulations.py
+++ b/sasdata/data_util/manipulations.py
@@ -65,18 +65,11 @@ def get_q_compo(dx: float, dy: float, detector_distance: float, wavelength: floa
 
 def flip_phi(phi: float) -> float:
     """
-    Force phi to be within the 0 <= to <= 2pi range by adding or subtracting
-    2pi as necessary
+    Force phi to be within the 0 <= to < 2pi range, using a modulus operation.
 
-    :return: phi in >=0 and <=2Pi
+    :return: phi in range 0 <= phi < 2pi
     """
-    if phi < 0:
-        phi_out = phi + (2 * math.pi)
-    elif phi > (2 * math.pi):
-        phi_out = phi - (2 * math.pi)
-    else:
-        phi_out = phi
-    return phi_out
+    return phi % (2 * np.pi)
 
 
 def get_pixel_fraction_square(x: float, x_min: float, x_max: float) -> float:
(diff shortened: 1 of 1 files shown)


ehewins:
Hmm. That's quite a lot of failed checks. Time to do some digging.

ehewins:
My last commit fixes the broken unit test, but I've discovered another bug which needs to be fixed. I think said bug has always been present, but was hidden by the one this PR set out to fix. The problem is this: you can't set `delta_phi [deg]` to 180 degrees (in the 'Edit Slicer Parameters' menu for Wedge Averaging in Phi), for the same reason I gave at the end of my previous comment. I have a feeling the solution is to be found in the code for the `Ring` class, since it is effectively doing the same job when the angles are set like this.

sasdata/data_util/manipulations.py
```diff
@@ -63,19 +63,18 @@ def get_q_compo(dx: float, dy: float, detector_distance: float, wavelength: floa
     return out
 
 
-def flip_phi(phi: float) -> float:
+def normalize_angle(phi: float, closed_at_end: bool = False) -> float:
     """
-    Force phi to be within the 0 <= to <= 2pi range by adding or subtracting
-    2pi as necessary
+    Force phi to be within the half-open interval 0 <= phi < 2pi, unless
+    closed_at_end is set to True, in which case the interval is 0 < phi <= 2pi.
```
krzywon, line 69: Why the difference and can you document those differences here?

krzywon (commented):
The fix works, but I could use a bit more understanding of where `closed_at_end` would be used and why.

sasdata/data_util/manipulations.py
```diff
@@ -63,19 +63,18 @@ def get_q_compo(dx: float, dy: float, detector_distance: float, wavelength: floa
...
-    if phi < 0:
-        phi_out = phi + (2 * math.pi)
-    elif phi > (2 * math.pi):
-        phi_out = phi - (2 * math.pi)
-    else:
-        phi_out = phi
+    phi_out = phi % (2 * np.pi)
+    if closed_at_end and phi_out == 0:
+        phi_out = 2 * np.pi
+
```
butlerpd, line 77: Clever way to replace the if/elif, if with a single line! However the closed at end seems clunky.  Am I correct in assuming the problem is that the modulus of 2pi/2pi is 0 is the problem? If so would using a `divmod(a,b)` operation and making the if statement be `if (b=0 and a>0) phi_out=2*np.pi` simplify things and not require the new `closed_at_end` variable which changes the calling interface to the function?

sasdata/data_util/manipulations.py
```diff
@@ -63,19 +63,18 @@ def get_q_compo(dx: float, dy: float, detector_distance: float, wavelength: floa
     return out
 
 
-def flip_phi(phi: float) -> float:
+def normalize_angle(phi: float, closed_at_end: bool = False) -> float:
```
butlerpd, line 66: I agree that the old nomenclature makes no sense.  I will admit to being too chicken myself to change it however. That is a brave move.   Are you absolutely sure than this function is not called from any other modules in the sasview code base?

sasdata/data_util/manipulations.py
```diff
@@ -303,13 +302,16 @@ def get_bin_index(self, value):
...
+            temp_x = math.log(value, self.base) - math.log(self.min, self.base)
             temp_y = math.log(self.max, self.base) - math.log(self.min, self.base)
         else:
-            temp_x = self.n_bins * (value - self.min)
+            temp_x = value - self.min
             temp_y = self.max - self.min
+        # Fixing an issue where certain angular coordinate values give bad bin values
+        if np.fabs(temp_x) > np.fabs(temp_y):
+            temp_x = normalize_angle(temp_x)
         # Bin index calulation
```
butlerpd, line 313: I don't think this is the appropriate fix is it?  this is a generalized binning class that returns a bin value given the min and max of a range, the number of bins being used, and a value whose bin is to be computed as I understand the class? In which case it should not know anything about the type of object that is being binned.  Thus `normalize_angle`, which assumes the range (min, max ) being binned is an angular range should not be in this class (for example boxslicer does not have angular ranges). If we need to trap such an error nicely maybe just return an error log message?  I would expect that the calling class making use of this binning class should be responsible to first call `normalize_angle` prior to sending to binning if that class is sending an angular range?

butlerpd (changes requested):

sasdata/data_util/manipulations.py
ehewins: I've double checked, and yes `flip_phi()` was only ever called within `manipulations.py`, and never in sasview. So this rename causes no problems.

sasdata/data_util/manipulations.py
ehewins: The class may not be aware of the type of object being binned, but based on how it's called I'm fairly certain the only time |`temp_x`| can ever be greater than |`temp_y`| would be when it's working with angular values. So it shouldn't cause problems for non-angular slicers unless the `value` parameter it's called with makes no sense (`value` would have to be greater than `self.max`). I see your point about ensuring the calling class normalises the angles instead, but for the bug this PR set out to solve, that isn't enough.

Let me explain the original issue a bit more thoroughly. For whatever confusing reason, `manipulations.py` is set up to have angles measured anticlockwise from the negative x-axis. So 0 - pi/2 is in the third quadrant, pi/2 - pi is in the fourth, pi - 3pi/2 is in the first quadrant and 3pi/2 - 2pi in the second. When `get_bin_index` is called for angular averaging, it is done in a loop with the `value` argument gradually moving in the anticlockwise direction, from `phi_min` to `phi_max`. Issue #40 occurs when `phi_min` is in the second quadrant, and `phi_max` is in the third. When `value` is in the second quadrant, `(value - min)/(max - min)` is between -1 and 0, weird but not problematic. When `value` is in the third quadrant however, `(value - min)/(max - min)` > 1. This is what results in the out of range bin indices. I tried experimenting with setting `min` and `max` in `Binning()` so that `max` is always greater, but it didn't help.

sasdata/data_util/manipulations.py
ehewins: I've implemented `divmod` as part of a further revision to this function.

sasdata/data_util/manipulations.py
ehewins: The `normalize_angle` method has been changed now, but I've supplied comments explaining the choices in the new version.

sasdata/data_util/manipulations.py
ehewins: I'm planning to do a full refactor of the binning methods used in `manipulations.py` once I've finished with these bugs, with the chief goal of introducing working fractional binning.  Ultimately I think circular/angular binning and square binning are different enough that they should be handled by different methods, but that can be saved for the re-work. For now, I'd like to keep this `normalize_angle()` call in place, especially since it isn't actually causing any issues for other types of averaging.

ehewins:
> My last commit fixes the broken unit test, but I've discovered another bug which needs to be fixed. I think said bug has always been present, but was hidden by the one this PR set out to fix. The problem is this: you can't set `delta_phi [deg]` to 180 degrees (in the 'Edit Slicer Parameters' menu for Wedge Averaging in Phi), for the same reason I gave at the end of my previous comment. I have a feeling the solution will be in the code for the `Ring` class, since it's effectively doing the same job when the angles are set like this.

I've noticed that this bug also applies to Wedge Averaging in Q. It seems you can't use the wedge slicer with a full circle ROI under any circumstances. Considering the problem mentioned in Issue #40 has been fixed here, I really ought to open a new issue and continue with the wedge slicer full-circle bug there.

sasdata/data_util/manipulations.py
butlerpd: I am confused by the suggestion that the binning class needs refactoring. It is actually a very general class that takes a range from x to y, a number of bins desired and a value you want to know the bin number of an returns it either in linear or log? I don't see how that is causing any problem? There are certainly questions I can imagine with dealing with how one weights a pixel that crosses a bin boundary but that is all done in the `get_pixel_fraction` function I thought? If not then it should be.

On the other hand, if the binning code DOES get thrown away, it seems to me that is an even stronger argument for putting the code where I believe it belongs: the sector class just before calling binning? That way your solution for dealing with sectors that straddle 0, whatever it is, is preserved and does not have to be redesigned into the new binning class?

Which brings me to my issue with the current fix, as suggested in my comment to issue #40, the error is not really coming from the `Binning` class. that class is doing what it should. the problem is that you are calling it with incorrect values (you are giving it a min > max basically).  That should be addressed it seems to me in the code that knows about that fact. I believe the place to address the problem is in the `_avg` method of the `_Sector` class.

In fact there is already a TON of code in that method dealing with **exactly** this straddling zero problem for other sector type slices.. and I mean a lot... as it should be I would argue (though see caveat at end)

Basically the fix should go around [line 889 ](https://github.com/SasView/sasdata/blob/40-return-correct-bin-index-for-angular-coordinates/sasdata/data_util/manipulations.py#L887-L894) where the binning object is created

```python
        #  set up the bins by creating a binning object
        if run.lower() == 'phi':
            binning = Binning(self.phi_min, self.phi_max, self.nbins, self.base)
        elif self.fold:
            binning = Binning(self.r_min, self.r_max, self.nbins, self.base)
        else:
            binning = Binning(-self.r_max, self.r_max, self.nbins, self.base)
```

and then in the averaging loop when retrieving the binning index aroun [line 950](https://github.com/SasView/sasdata/blob/40-return-correct-bin-index-for-angular-coordinates/sasdata/data_util/manipulations.py#L950-L955) 

```python
            # Get the binning index
            if run.lower() == 'phi':
                i_bin = binning.get_bin_index(phi_value)
            else:
                i_bin = binning.get_bin_index(q_value)

            # Take care of the edge case at phi = 2pi.
            if i_bin == self.nbins:
                i_bin = self.nbins - 1
```
Unfortunately looking at the current code I have realized this evening that I'm pretty sure that math is wrong. everything will be put in the lower index bins unless I'm doing the math wrong? I think the easiest way to deal with this correctly if one wants to stick to the 0<phi<2pi  range that all this code seems to like, would be to implement a rotation of the wedge such that phi_min is 0.  Speaking with @lucas-wilkins just now, assuming we don't care about 0<phi<2pi then rotating the wedge by 2pi would be even easier to implement code wise (just add 2pi to each of phi_min, phi_max, and phi_value).  given the transformation is just to send to binning I would agree that there is no need to try to keep 0<phi<2pi. 

In fact it is even simpler I realize, which @lucas-wilkins may have already known. All you need to do is add 2pi to phi_max if phi_min > phi_max and add 2pi to phi_value if phi_min > phi_value 

so my thinking is that the end result would look something like this?

on line 889 add
```phython
if phi_min > phi_max: phi_max =+ 2*np.py
```
to yield
```python
        #  set up the bins by creating a binning object
        if run.lower() == 'phi':
            if phi_min > phi_max: phi_max =+ 2*np.py
            binning = Binning(self.phi_min, self.phi_max, self.nbins, self.base)
        elif self.fold:
            binning = Binning(self.r_min, self.r_max, self.nbins, self.base)
        else:
            binning = Binning(-self.r_max, self.r_max, self.nbins, self.base)
```
and on line 953 (was 952) add
```python
if phi_min>phi_value: phi_value =+2*np.py
```
```python
            # Get the binning index
            if run.lower() == 'phi':
                if phi_min>phi_value: phi_value =+2*np.py
                i_bin = binning.get_bin_index(phi_value)
            else:
                i_bin = binning.get_bin_index(q_value)

            # Take care of the edge case at phi = 2pi.
            if i_bin == self.nbins:
                i_bin = self.nbins - 1
```
Having now gone through this whole exercise, I'm wondering if this method would not allow one to clean up a bunch of the lines of code currently trying to deal with this problem?

butlerpd: Just to be clear - that last bit is way out of scope for this PR. Just an observation that may be worth another look at some point?

ehewins: When I talk about refactoring the binning, I don't just mean refactoring the class `Binning`, I mean a redesign of how bins work everywhere in `manipulations.py` to get fractional binning working. Somebody has started working on this with `get_pixel_fraction`, but it's not implemented yet. So, I'll be spending a lot more time thinking about these problems in the near future.

I think you're right that it shouldn't be `Binning.get_bin_index`'s job to sort out this zero-straddling error. I wanted to get a quick fix in so that wedge slicer could be made fully functional, and the approach I've gone with so far was just the first thing I found that worked. I'll try out the fix you've suggested, but the commit history in this branch is getting pretty messy. Perhaps I should just make a new branch addressing Issue #40 and scrap this one.

ehewins: I've created a new branch and new PR which implements an alternative fix which I think everyone involved will be much happier with. You can find it here: #42

PR #42, opened 2023-08-04
Fixing Issue #40 (but properly this time)

### LINKED ISSUE #40: SectorPhi class throws errors (bin index out of range) when phi range includes 0 (or 2Pi)

ehewins:
This PR is an alternative approach to fix #40 
My original attempt at fixing this issue can be found in PR #41 , but I've opened a new branch and new PR to avoid making a mess of the commit history. The old approach worked, but as @butlerpd pointed out, it didn't go about it in the best way. The new approach found here is based on Paul's suggestions and solves the problem without making the `Binning.get_bin_index()` method do more than it's supposed to.

Fixes #40

### DIFF
--- a/sasdata/data_util/manipulations.py
+++ b/sasdata/data_util/manipulations.py
@@ -881,9 +881,20 @@ class _Sector:
 
         #  set up the bins by creating a binning object
         if run.lower() == 'phi':
-            if self.phi_min > self.phi_max:
-                self.phi_max += 2 * np.pi
-            binning = Binning(self.phi_min, self.phi_max, self.nbins, self.base)
+            # self.phi_min/max are waht are received from the GUI and can be in the range of 0, 2pi;
+            # -delta, 2pi-delta; or delta, 2pi+delta, while phi_min/max move the received values to
+            # always in the range 0, 2pi. Use phi_min/max here to ensure well understood behavior
+            # (though, if careful, one could use the self.phi_min/max.
+            # NOTE: if using phi_min/max as here instead of self.phi_min we must NOT change
+            # their values as they are used throughout to determine what points (also all in
+            # the 0, 2pi range) are in the ROI.
+            # NOTE2: the first line is needed in any case to ensure that when a range straddles
+            # the discontinuity inherent in circular angles (e.g. jumping from 2pi to 0) by
+            # putting sliding the range to be continuous (e.g. 2pi goes to 2pi + delta)
+            if phi_min > phi_max:
+                binning = Binning(phi_min, phi_max + 2 * np.pi, self.nbins, self.base)
+            else:
+                binning = Binning(phi_min, phi_max, self.nbins, self.base)
         elif self.fold:
             binning = Binning(self.r_min, self.r_max, self.nbins, self.base)
         else:
@@ -946,11 +957,15 @@ class _Sector:
 
             # Get the binning index
             if run.lower() == 'phi':
-                if phi_value < binning.min:
-                    phi_value += 2 * np.pi
-                elif phi_value > binning.max:
-                    phi_value -= 2 * np.pi
-                i_bin = binning.get_bin_index(phi_value)
+                # phi_value is in the range of 0 to 2pi. If the original range had to be slid
+                # to accommodate straddling the discontinuity in the circular angles
+                # (e.g. jumping from 2pi to 0), we need to slide the range for phi_value
+                # in the same way so that it falls in the continuous range set up for the
+                # binning process
+                if phi_min > phi_value:
+                    i_bin = binning.get_bin_index(phi_value+ 2* np.pi)
+                else:
+                    i_bin = binning.get_bin_index(phi_value)
             else:
                 i_bin = binning.get_bin_index(q_value)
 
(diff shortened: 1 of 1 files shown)


sasdata/data_util/manipulations.py
```diff
@@ -944,6 +946,10 @@ def _agv(self, data2D, run='phi'):
 
             # Get the binning index
             if run.lower() == 'phi':
+                if phi_value < binning.min:
+                    phi_value += 2 * np.pi
+                elif phi_value > binning.max:
+                    phi_value -= 2 * np.pi
```
butlerpd, line 952: Are you sure this is correct?  I think this should be omitted as I understand the math? Maybe we can discuss? I could be doing the maths wrong. As @lucas-wilkins points out : "angles are the WORST" 

my thinking:
for example if phi_min is 320 degrees, phi_value is 340 and phi_max (on the other side of 0) is 20.  In this case phi_value is > phi_min which is correct. But it is also > phi_max.  I don't see why you need to subtrac 2pi which just makes it a negative angle ... and would now make phi_value < phi_min which violates condition one?

butlerpd (changes requested):

sasdata/data_util/manipulations.py
ehewins: It must be correct. Removing these lines causes the bug to return (it's fixed when the central line is above the x-axis, but still broken when below it). `self.phi_min` and `self.phi_max` aren't restricted to the 0 - 2pi range, and the check on line 884 ensures self.phi_max is always numerically greater than self.phi_min. Therefore, if `phi_value` > `binning.max` (which is the same as `self.phi_max`), then this subtraction has to take place or `phi_value` will be outside the allowed range and `get_bin_index` will return a value > `self.nbins`.

butlerpd: Now I am very (very) confused. Of course not failing <> from correct, but still, I cannot even see how that second bit of code ever gets used since phi_value at that point should always be smaller than phi_max no matter what (If we've straddled zero phi_max should be >2PI while phi_value should not be)? Unless wedge_slicer is sending something over I don't understand? But if the code fails without that `elif` statement then it clearly is hitting that code. Going to have to crack my geometry books here.

butlerpd: Right.... so we may both be wrong. I'm starting to feel like an idiot 😒 .  I think I'm beginning to see why they created `flip_phi`. I believe that most of the problem goes away without adding these tests, if done right. But there is still a few ends I've not quite finished figuring out which I think come from `WedgeInteractor` maybe?  @lucas-wilkins is right: angles are a real pain 😕.

Will continue tomorrow and hopefully post my findings.

ehewins: > we may both be wrong

I'm don't actually understand what the problem is here. There must be something you've got in mind which hasn't occurred to me, because as far as I can see this code is fine.

butlerpd: Ok @lucas-wilkins and @ehewins, I’ve now done what I had avoided doing so far: really understand how the sector class works in `manipulation.py`.  Changes needed are of course minor but not as simplistic as I’d originally thought ☹️. The really tricky/confusing bit is that there are like 4(?) Different angular range spaces (-pi, pi; 0,2pi; -delta, 2pi-delta; 0+delta,2pi+delta). `phi_val` and `phi_max/min` are in the 0, 2pi interval. However `self.phi_min/max` (yes `self.phi` is different from `phi` ☹️) is unknown and can be in any of the last 3 ranges.  The gui slicer is in the -pi to pi range natively.
Trying to understand it all I ended up writing my own version of manipulations.py changes which is very close to  yours @ehewins , but not the same. I have created a new branch off of your branch (called v3) as it was the easiest way I could think of to show what I did and why -- which also means it has extensive comments which would be good to capture somewhere 🙂 The difference is that, having figured out the subtle differences in angular intervals, I tried to keep them separate AND not change that range like I originally did (e.g. using phi_max += 2* np.py) to try to keep what I was doing clean and clear (else I confuse myself). 
Now that I understand it, I believe your code also is correct. The main difference is that you are mixing `phi_value` (in 0, 2pi space) with `self.phi_min/max` (=`binning.min/max`) which can be in any of 3 spaces (in fact 2 I think since you actually altered `self.phi_max`).  That is why you ended up with the `if elif` that confused me. They are both required in order to deal both with the straddling problem and the fact that `phi_value` might be in a different angular space. The other difference is just a preference of starting point.  My plot in sasview gui doesn’t go negative till phi_min is below the x axis (so the plot can go more than pi but never less than -pi) while your version will start negative when the central line crosses the x axis meaning that the plot can go to less than -pi but not more than pi.
To answer @lucas-wilkins  point about tests, I don’t know of any good test to verify that we are including all the points we are supposed to from the ROI, but we can do something close if we had some rocking curve of flux line data I think. We also would need to have tests for each of 3 cases: not straddling the discontinuity, straddling but with the central line above the x axis and the central line below the x axis.
While it is really important to get some good unit tests, I think that should not stop this PR as long as a new issue is put in for that?

butlerpd:
I suggest you look at my version with documentation and test. I'm not sure if my version would be easier or not for a new developer 3 years from now to pick up (modulo the documentation some version of which should be included here if this is the version we go with).  I'm happy now to approve either way. As I say in the conversation above, unit tests will be important but will take time to get the appropriate data and write so I suggest we DON'T wait for that.

butlerpd:
I have one other question I completely missed. I do not know phyton or OOP well enough to know the answer.  Your code as written changes the value of `self.phi_max`.  Is `self.phi_max` a purely local variable? or are you making the `_Sector` class in sasdata change a parameter value that belongs to sasview gui?

If it is local I'm fine. If however we would be changing a `WedgeSlicer.py` value behind its back so to speak then I think we need to not do that! That would be very bad form I think. In principle it should be ok (2pi differences in angles), but you are changing it here precisely because it does matter in this case so I fear unintended consequences in the future .. particularly by Ellis 2 and beyond 😄.  At the least it would probably lead to unnecessary confusion for the next person trying to work in this part of the code base. The fix of course would be trivial -- use the same construct as in version 3 of this branch (passing an expression containing `self.phi_max` rather than an altered value of `self.phi_max`.  You would also then need to replace `self.phi_max` with `binning.max` in the calculation of `step` (for clarity probably best in that case to also replace `self.phi_min` with `binning.min`)

butlerpd (approved):

butlerpd:
OK - in that case I'm good with it.  I would like to see documentation at least in comments before a final merge? Ultimately it would probably be best to move a lot of that into the class doc strings?

PR #53, opened 2023-09-29
29: Building sasdata documentation

### LINKED ISSUE #29: Move documentation from sasview
The documentation for the sasdata package is still housed in sasview. That should all be moved here along with the sphinx infrastructure to build and bundle it.

A revisit to the documentation wouldn't hurt during the process.

krzywon:
This is the first step to including the sasdata documentation in remote repositories (refer to sasview/sasview#2641). This auto-generates the docs during the CI process, but they are not currently stored anywhere. The next step will be to bundle the docs with the pypi release.

Much of the user documentation was ported from the SasView docs and modified for this stand-alone package.

Fixes #29


wpotrzebowski:
Should be part of 6.0 release

llimeht (approved):

.github/workflows/test.yml
```diff
@@ -69,9 +69,9 @@ jobs:
     - name: Build sasdata docs (Ubuntu only)
       if: matrix.os == 'ubuntu-latest'
       run: |
-        mkdir -p build/lib
-        sphinx-apidoc -o docs/source/dev/generated/ -d 8 -H SasData build/lib
-        sphinx-build -M html docs/source/ docs/build/
+        cd docs
```
llimeht, line 72: The three lines can be replaced with `make -C docs html` — let make deal with the working directory for us.

docs/source/user/data/data_formats_help.rst
```diff
@@ -0,0 +1,217 @@
...
+
+In essence, coordinate formats specify a location and one or more properties of
+that location (e.g. what it represents, its volume, or magnetisation, etc). The
+PDB/OMF/SLD formats all use a rectangular grid of voxels.
+
+The .STL coordinate format is not currently supported by SasView.
+
+**Examples of these formats can be found in the \\example_data\\coordinate_data
+sub-folder in the SasData distribution.**
+
```
wpotrzebowski, line 150: I may be wrong but isn't VTK format missing for GSC?

wpotrzebowski (approved):
Code looks good. Some improvements suggested. I didn't test functionality/build though

docs/source/user/data/data_formats_help.rst
krzywon: It's questionable if this should even be in the sasdata documentation while the coordinate data readers and data objects are still in sasview (see https://github.com/SasView/sasview/issues/2656).

butlerpd: Good point but to @wpotrzebowski point, I thought GSC is in fact where VTK is being used by Dirk for his magnetic stuff (but only a  specific version?)

wpotrzebowski:
@smk did you have a chance to look at this PR?

smk78:
@krzywon, looking at the docs generated by https://github.com/SasView/sasview/actions/runs/7106711492:

I see the addition of a sasdata entry in the Developer Documentation Contents:

image

This leads to a single entry list:

image

Which, in turn, leads to a dead-end:

image

Looking at the files changed by the PR one can see that two new docs have been created, _Importing Data_ and _Exporting Data_. These are present:

image

But are only discoverable using the search, they are not linked from anywhere.

PR #56, opened 2023-12-06
55: Generate Developer Docs

### LINKED ISSUE #55: Developer Documentation Not Building Properly
As noted by @smk78 in https://github.com/SasView/sasdata/pull/53#issuecomment-1842987694, the developer documentation is not being built properly. After some investigation, this occurred with the change in https://github.com/SasView/sasdata/pull/53/commits/708f332f99f6cd3331a82afc1d5a59e692f377d8. The `sphinx-apidoc` and `sphinx-build` commands need to be reworked to ensure they are formatted properly.

krzywon:
As noted by @smk78, the developer documentation was not present in the build. This PR corrects that mistake.

To check the docs are bundling with sasview properly, please check the builds in https://github.com/SasView/sasview/actions/runs/7121411183/

Fixes #55

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,14 +1,17 @@
 # Minimal makefile for Sphinx documentation
-#
 
 # Environment variables
 SPHINXOPTS    ?=
 SPHINXBUILD   ?= sphinx-build
 SPHINXAPIDOC  ?= sphinx-apidoc
 APIOPTS       = -d 8 -H SasData
+DOCSDIR       = docs
 SOURCEDIR     = source
 BUILDDIR      = build
-CD            = cd
+DEVDIR        = dev
+LIBDIR        = lib
+GENDIR        = generated
+UP            = ..
 
 ifdef ComSpec
     RMDIR   = rmdir /s/q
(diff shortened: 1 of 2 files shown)


smk78:
Tested https://github.com/SasView/sasview/actions/runs/7121411183/ on W10/x64.

This PR has indeed fixed the broken sasdata API doc build.

However, it has not addressed a couple of things from my review on #53 .

1) When you go to Developer Documentation you see this:
image
which would be fine, if the link to _sasdata overview_ did give an overview in the same way _sasmodels overview_ does. But all it does is link to this:
image
which, in turn, links to API docs:
image
etc.
So my question here is what exactly is the role of /doc/dev/sasdata-dev/dev.html? It seems to be redundant? In which case, why not do away with it and change the top level link from _sasdata overview_ to _sasdata API_?

2) The documents /doc/user/data/data/data_import_help.html and /doc/user/data/data/data_export_help.html are still not linked from anywhere (not even from the 'API' docs) and are only accessible through the search box.

smk78 (changes requested):
Looking good. The API doc build is now fixed, but there are still a couple of oddities. See above.

krzywon:
@smk78 - I agree with your first point and just pushed a change to remove the unneeded documentation level.

As far as your 2nd point, the `data_import` and `data_export` help files are specific to the sasdata scripting interface. All of the same information is available in the sasdata developer docs. I purposely omitted those two files from the SasView docs, because the information isn't relevant to the users of the GUI-based program.

smk78 (approved):
Oh, I see. Ok, then this should be good to merge.

butlerpd:
> As far as your 2nd point, the `data_import` and `data_export` help files are specific to the sasdata scripting interface. All of the same information is available in the sasdata developer docs. I purposely omitted those two files from the SasView docs, because the information isn't relevant to the users of the GUI-based program.

So reading @smk78 comment -- does this mean these files are orphans because they are duplicates? in which case should we delete them? or are you saying those files are linked through the `sasdata` developer docs?  I worry about "same information" available elsewhere as it makes maintenance a nightmare 😄.

PR #58, opened 2024-01-18
Update example data

smk78:
This PR updates the example data library in sasdata to match that in sasview main. This includes the supplied but not committed magnetic sans data from https://github.com/SasView/tutorials/issues/13.

testdata_help.rst has also been revised accordingly.


krzywon:
@smk78 - The majority of these data sets were purposely omitted during the initial move. The magnetic data directory was overlooked, but coordinate data and save states are only importable through SasView.

Coordinate data readers and data structures should be here, but are not, as of yet: sasview/sasview#2656

Save states could also live here in the future, but that would require this package to load the data.

smk78:
> Good point. I guess we need to think hard about what package loads what data but then we want to be able to list all of them in one place for the full SasView installer I gues? But for now this is a first step I guess?

So some thought will be needed about testdata_help.rst.  This presently covers all our test/example data (ie, including those files not in or supported by sasdata).

butlerpd:
This looks like it should also be merged into 0.9 eventually so it will get bundled with 6.0?

butlerpd:
@krzywon probably needs to verify that this is now appropriate?

krzywon (approved):
This looks better now. I have a small issue with the help file (it still mentions SasView specific files), but I don't think that should hold this PR up

smk78:
> This looks better now. I have a small issue with the help file (it still mentions SasView specific files), but I don't think that should hold this PR up

When we release what example data will actually be included:
 i) only that in sasview (ie, all our example data as at present);
 ii) only that in sasdata (ie, no coordinate_data or saved states);
 iii) some clever synthesis of what's in sasview & sasdata so that everything is included but is dragged from the different repos?

I'm presuming not i).
ii) concerns me; it would render the GSC help docs useless for a start.
iii) would be ok if we come up with a way of splitting the testdata_help between the repos. But it's not obvious to me how to do that neatly.

Issue #57, opened 2024-01-18
Failing Unit Tests: MacOS and Ubuntu


krzywon:
Two unit tests are failing for both MacOS and Ubuntu, across all versions of Python. These tests are successful under Windows.

Link to failing job: https://github.com/SasView/sasdata/actions/runs/7573626662/job/20626178693

Error Message:

> test/sasdataloader/utest_extension_registry.py:95: AssertionError
=========================== short test summary info ============================
FAILED test/sasdataloader/utest_cansas.py::cansas_reader_xml::test_invalid_cansas - AssertionError: False is not true
FAILED test/sasdataloader/utest_extension_registry.py::ExtensionRegistryTests::test_data_reader_exception - AssertionError: 4 != 1
========================= 2 failed, 98 passed in 6.50s =========================

krzywon:
A quick investigation suggests changes made in the lxml library are the cause of the unit test failures. More investigation is required. In the mean time, a PR has been submitted to limit lxml to versions less than 5.

llimeht:
As an additional data point, the test suite from sasdata version 0.8.1 passed on Debian's CI with python3-lxml version 5.1.0-1. Have these tests or those code paths changed since then? 
https://ci.debian.net/packages/s/sasdata/testing/amd64/41845113/

krzywon:
This issue has been resolved, but only for the short-term. #64 is more specific to the long-term issue so I am closing this.

PR #63, opened 2024-01-24
57: Restrict lxml to versions less than 5.0

krzywon:
This limits the version of the XML reader library, [lxml](https://lxml.de/), to under v5.0. The changes made in v5.0 and beyond seem to break unit tests in MacOS and Ubuntu, but not for Windows for some reason. This is a temporary fix until a proper solution can be investigated.

Refs #57 (NOTE - this should not close the issue)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 # Data import/export requirements
 h5py
-lxml
+lxml < 5.0
 
 # Calculation
 numpy


butlerpd:
Should this also be pushed to branch 0.9 that gets bundled with sasview 6.0?

butlerpd (approved):
Certainly seems to remove the problem and CI is otherwise failing.  The only Caveat is we should  make investigating the root cause a priority or we will forget and get stuck down the road?

In the meantime I recommend merging post hast

PR #79, opened 2024-08-06
Refactoring Branch

lucas-wilkins:
Main branch for the refactoring project


sasdata/quantities/units.py
```diff
@@ -0,0 +1,3415 @@
...
+        "hr": hours,
+        "hour": hours,
+        "a.u.": atomic_mass_units,
+        "amu": atomic_mass_units,
+        "degr": degrees,
+}
+
+
+#
+# Units by type
```
jamescrake-merani, line 1959: While writing the unit tests for my parser, I had a thought. I think the unit groups should perhaps go into a separate module. To import a unit, we would do

```python
from sasdata.quantities.units import meters. meters_per_second ...'
```

Which makes sense. But to import a unit group, the import statement would look like this.

```python
from sasdata.quantities.units import speed, acceleration ...'
```

Which might be a little confusing because we're importing a unit *group* from the units module. Its a fairly minor change but having a separate unit group module might make this clearer.

lucas-wilkins: Yeah, sounds like a good plan

sasdata/transforms/operation.py
```diff
@@ -0,0 +1,19 @@
+import numpy as np
+from sasdata.quantities.quantity import Quantity
```
jamescrake-merani, line 2: Can this file be deleted now? Seems to just be a sketch which has now been implemented in `operations.py`.

sasdata/temp_hdf5_reader.py
```diff
@@ -0,0 +1,154 @@
...
+        for root_key in f.keys():
+
+            entry = f[root_key]
+
+            data_contents = []
+            raw_metadata = {}
+
+            entry_keys = [key for key in entry.keys()]
+
+            if "sasdata" not in entry_keys and "data" not in entry_keys:
```
jamescrake-merani, line 124: Does this account for the situation where `entry_keys` contains something like `'sasdata01'`?

lucas-wilkins: nope

PR #103, opened 2024-10-28
Move to numpy 2+

### LINKED ISSUE #72: numpy and scipy have breaking changes
As with sasmodels and sasview, the sasdata code base needs to be updated for the breaking changes in `numpy 2.0.0` and `scipy 1.14`. In particular NaN and NAN are no longer valid in numpy.  Currently we are pinning to an older version of numpy and scipy for release 6.0 but this should not be forgotten

krzywon:
This should make sasdata numpy 2+ compatible. This will need to be merged prior to any sasview changes.

Fixes #72

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ h5py
 lxml
 
 # Calculation
-numpy==1.*
+numpy
 
 # Unit testing
 pytest
--- a/sasdata/data_util/formatnum.py
+++ b/sasdata/data_util/formatnum.py
@@ -278,10 +278,10 @@ def test_compact():
     # non-finite values
     assert value_str(-np.inf,None) == "-inf"
     assert value_str(np.inf,None) == "inf"
-    assert value_str(np.NaN,None) == "NaN"
+    assert value_str(np.nan,None) == "NaN"
     
     # bad or missing uncertainty
-    assert value_str(-1.23567,np.NaN) == "-1.23567"
+    assert value_str(-1.23567,np.nan) == "-1.23567"
     assert value_str(-1.23567,-np.inf) == "-1.23567"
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -3,7 +3,7 @@ h5py
 lxml
 
 # Calculation
-numpy==1.*
+numpy
```
jamescrake-merani, line 6: Should this be `numpy==2.*` to ensure the user has numpy 2 installed, and not 1 (or a future version)?

jamescrake-merani (approved):
LGTM. Only thing is: in the `_format_uncertainty` function in `formatnum.py`, should this still be returning `NaN` with the capitals, or is this unaffected?

krzywon:
> LGTM. Only thing is: in the `_format_uncertainty` function in `formatnum.py`, should this still be returning `NaN` with the capitals, or is this unaffected?

This is only used to generate a string representation of the uncertainties. Capitalization shouldn't matter.

requirements.txt
krzywon: If numpy versions properly, v1.x releases should never be the 'latest'.

bmaranville: No, the changes made here seem to be backward compatible with numpy < 2... `numpy.NaN` used to be an alias for `numpy.nan` in numpy 1.2x, and `numpy.string_` was an alias for `numpy.bytes_`.  This PR just removes the use of those deprecated aliases.  You might want to set a minimum version of 1.20 or something.

jamescrake-merani: > If numpy versions properly, v1.x releases should never be the 'latest'.

I was more thinking of if the user had Numpy 1 already installed, it would satisfy this requirement. But if the changes are backwards compatible then that won't be an issue.

PR #104, opened 2024-12-05
Fix small error in the refactor branch

jamescrake-merani:
The summary of metadata was printing the title twice: once where its meant to be, and once for the definition when really the definition should've been printed.


lucas-wilkins:
Is this still valid?

jamescrake-merani:
I'm not sure as this may have been overwritten by the changes @rprospero made. If so, we can close this PR.

PR #106, opened 2025-01-30
Sasview database

summerhenson:
Database backend in progress.


sasdata/fair_database/data/tests.py
```diff
@@ -0,0 +1,123 @@
...
+from django.contrib.auth.models import User
+from rest_framework.test import APIClient, APITestCase
+from rest_framework import status
+
+from .models import Data
+
+def find(filename):
+    return os.path.join(os.path.dirname(__file__), "../../example_data/1d_data", filename)
+
+class TestLists(TestCase):
```
krzywon, line 15: Other suggested tests:

- Modify data using PUT (public data while authenticated and not, private data while authenticated as data owner, authenticated as other user, and unauthenticated)
- DELETE public and private data (all user cases listed in last bullet) - Should public data be deletable? By the administrator, yes. By a random user? That's a question I'll think about...

krzywon (commented):
In addition to my inline comments, a few general notes:

- I said this in the meeting, but wanted to reiterate it - good use of atomic commits with specific commit messages
- Some python formatting would be useful. Running everything through ruff could be worthwhile.
- Relative imports can be fragile, but mostly when trying to run a python file directly. That isn't easy with Django, so it's not that huge an issue, but wanted to mention it.
- Another thing we discussed at today's meeting, but wanted to mention again, is user auth and request.user clashes. Many of the checks don't ensure the authenticated user is the data owner.

Overall, a good start to the project. I'll try to make comments every week, or as updates are made.

sasdata/fair_database/fair_database/permissions.py
```diff
@@ -0,0 +1,29 @@
...
+
+def is_owner(request, obj):
+    return request.user.is_authenticated and request.user == obj.current_user
+
+
+def has_access(request, obj):
+    return (
+        is_owner(request, obj)
+        or request.user.is_authenticated
+        and request.user in obj.users.all()
```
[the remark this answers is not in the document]
krzywon: Side comment to this - Explicit is usually better than Implicit, so I would suggest you wrap logical groups within parenthases.

sasdata/fair_database/user_app/tests.py
```diff
@@ -0,0 +1,158 @@
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response2.status_code, status.HTTP_200_OK)
+
+    # Test simultaneous login by multiple clients
+    def test_multiple_login(self):
+        response = self.client1.post("/auth/login/", data=self.login_data_2)
+        response2 = self.client2.post("/auth/login/", data=self.login_data_2)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response2.status_code, status.HTTP_200_OK)
+        self.assertNotEqual(response.content, response2.content)
```
krzywon, line 62: Checking if the tokens remain valid after two logins would also be a good check

sasdata/quantities/_units_base.py
```diff
@@ -265,6 +276,12 @@ def __repr__(self):
     def parse(unit_string: str) -> "Unit":
         pass
 
+    def serialise_json(self):
```
krzywon, line 279: Will this get extended to Child classes as well?

sasdata/quantities/quantity.py
```diff
@@ -995,6 +996,12 @@ def hash_data_via_numpy(*data: ArrayLike):
 
 QuantityType = TypeVar("QuantityType")
 
+# TODO: figure out how to handle np.ndarray serialization (save as file or otherwise)
```
krzywon, line 999: [pickle.dumps](https://docs.python.org/3/library/pickle.html) or [numpy.save](http://docs.scipy.org/doc/numpy/reference/generated/numpy.save.html)

sasdata/quantities/quantity.py
```diff
@@ -1013,7 +1020,11 @@ def jacobian(self) -> list[Operation]:
         return [self.operation_tree.derivative(key) for key in self.reference_key_list]
 
     def _recalculate(self):
-        """ Recalculate the value of this object - primary use case is for testing """
+        """ Recalculate the value of this object - primary use case is￼
+￼
+￼
+￼
```
krzywon, line 1026: ?

sasdata/quantities/_units_base.py
summerhenson: I believe that was the idea, but on a second look it might make more sense to just use the string representation of a unit for serialization.

sasdata/quantities/quantity.py
summerhenson: Probably a typo

sasdata/quantities/quantity.py
summerhenson: Wouldn't there be a security issue with pickle?

sasdata/fair_database/data/test/test_datafile.py
```diff
@@ -0,0 +1,380 @@
...
+        cls.user = User.objects.create_user(
+            username="testUser", password="secret", id=1
+        )
+        cls.data = DataFile.objects.create(
+            id=1, current_user=cls.user, file_name="cyl_400_20.txt", is_public=False
+        )
+        cls.data.file.save("cyl_400_20.txt", open(find("cyl_400_20.txt"), "rb"))
+        cls.client1 = APIClient()
+        cls.client1.force_authenticate(user=cls.user)
+        cls.client2 = APIClient()
```
backmari, line 108: The tests look great. I like that they are small and that the names show what they test. Here I would suggest renaming `client1` to `client_authenticated` and `client2` to `client_unauthenticated` to make the intent clear.

sasdata/fair_database/data/models.py
```diff
@@ -59,8 +57,9 @@ class DataSet(Data):
     )
 
     # data contents - maybe ManyToManyField
-    data_contents = models.ManyToManyField("Quantity", through="LabeledQuantity")
+    data_contents = models.ManyToManyField("Quantity")
```
backmari, line 60: Maybe I am missing something, but to reduce complexity, I would suggest starting with `data_contents` as a `OneToManyField`. We can reduce storage by allowing a quantity to belong to multiple data sets, but I think it will be cleaner and simpler to manage if every data set has its own data contents.

sasdata/fair_database/data/models.py
```diff
@@ -166,26 +164,26 @@ class OperationTree(models.Model):
...
-'''
 class Session(Data):
     """Database model for a project save state."""
 
+    # title
+    title = models.CharField(max_length=200)
+
     # dataset
-    # dataset = models.ManyToManyField(DataSet)
+    dataset = models.ManyToManyField(DataSet)
```
backmari, line 174: Same question here, should we for simplicity say that a data set can only belong to one session? These are things we can discuss.

sasdata/fair_database/data/models.py
```diff
@@ -87,6 +87,10 @@ class Quantity(models.Model):
 
     label = models.CharField(max_length=50)
 
+    dataset = models.ForeignKey(
+        DataSet, on_delete=models.CASCADE, related_name="data_contents"
+    )
+
```
backmari, line 93: This is the right way to connect Quantity to DataSet and better than my suggestion :ok_hand:

sasdata/fair_database/data/models.py
```diff
@@ -87,6 +87,10 @@ class Quantity(models.Model):
```
backmari, line 87: I saw your comment/question about how deleting an OperationTree should work. I think cascade delete would handle this automatically if OperationTree had Quantity as an optional foreign key with `on_delete=models.CASCADE`, and the same for the foreign key to the parent operation.

krzywon (commented):
Some more nitpicking 😉

sasdata/fair_database/data/models.py
```diff
@@ -0,0 +1,189 @@
...
+    """Database model for a set of data and associated metadata."""
+
+    # dataset name
+    name = models.CharField(max_length=200)
+
+    # associated files
+    files = models.ManyToManyField(DataFile)
+
+    # metadata - maybe a foreign key?
+    # TODO: when MetaData is finished, set blank/null false
```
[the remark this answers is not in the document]
summerhenson: The OneToOneField got moved to metadata, but metadata is a required field for creating a dataset

sasdata/fair_database/data/models.py
```diff
@@ -0,0 +1,210 @@
...
+
+    session = models.ForeignKey(
+        "Session",
+        on_delete=models.CASCADE,
+        related_name="datasets",
+        blank=True,
+        null=True,
+    )
+
+    # TODO: update based on SasData class in data.py
```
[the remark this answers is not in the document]
summerhenson: DataSetType is labeled as a rough draft in dataset_types.py, so I've opted to leave it out as a field in DataSet for now, but that should be revisited as the refactor progresses, as well as potentially other aspects of the SasData class should they change. The TODO is mostly a note for other people, unless there's a better way of indicating what will need another look later

backmari:
If there are no more planned changes to the models, maybe we should delete the migrations and create new initial migrations based on the current state of the models. It definitely made sense to keep all intermediate migrations while the models were being developed, but this is probably a good place to start fresh.

krzywon:
Due to the conflicts in this branch, I've recreated the work here in #213 and I am closing this PR. @summerhenson will still be credited with the commits.

PR #107, opened 2025-02-05
Cubic interpolation

rprospero:
This pr extends `calculate_interpolation_matrix_1d` to support cubic interpolation.  I've also duplicated the linear interpolation tests to also test the cubic interpolation.  Those tests were a significant helpful at catching a boundary condition bug that popped up when I vectorized the code (solved by the `if` statement on line 184).


sasdata/transforms/test_interpolation.py
```diff
@@ -88,4 +88,77 @@ def test_linearity_linear():
     linear_points = x_and_y @ mapping
 
     for t, e in zip(new_x.in_si(), linear_points.in_si()):
```
jamescrake-merani, line 90: Not really a change made in this branch but this file should be in the test directory (at the root of the repository). This might be something we move later in the refactor branch but it needs to be moved at some point because iirc the CI test runner is not running this test.

sasdata/transforms/test_interpolation.py
```diff
@@ -88,4 +88,77 @@ def test_linearity_linear():
...
+    for t, e in zip(y_values_test, y_values_expected):
+        assert t == pytest.approx(e, abs=2)
+
+
+@pytest.mark.parametrize("fun", test_functions)
+def test_cubic_interpolate_different_units(fun: Callable[[Quantity[ArrayLike]], Quantity[ArrayLike]]):
+    original_points = NamedQuantity("x_base", np.linspace(-10,10, 107), units.meters)
+    test_points = NamedQuantity("x_test", np.linspace(-5000, 5000, 11), units.millimeters)
+
+    mapping, _ = calculate_interpolation_matrix_1d(original_points, test_points, order=InterpolationOptions.CUBIC)
```
jamescrake-merani, line 129: Could this logic be abstracted into another function? Both `test_cubic_interpolate_matrix_inside` , and `test_cubic_interpolate_different_units` seem to have a lot of repeated logic.

sasdata/transforms/test_interpolation.py
```diff
@@ -88,4 +88,77 @@ def test_linearity_linear():
...
+    y_original = fun(original_points)
+    y_test = y_original @ mapping
+    y_expected = fun(test_points)
+
+    test_units = y_expected.units
+
+    y_values_test = y_test.in_units_of(test_units)
+    y_values_expected = y_expected.in_units_of(test_units)
+    #
+    # print(y_values_test)
```
jamescrake-merani, line 140: Does this comment still need to be here? Or if it does, is there perhaps a way to create an if statement so it only makes the plots under a certain condition?

jamescrake-merani (changes requested):
I haven't verified the maths but overall the code looks good to me. Please see the comments I've made, and let me know what you think.

sasdata/transforms/test_interpolation.py
jamescrake-merani: If you want, you can resolve this straight away but lets make sure we make this change once this has been merged.

sasdata/transforms/test_interpolation.py
rprospero: I've reworked the tests to parameterize over the different interpolation types, so that gets rid of all of the boilerplate.

jamescrake-merani: Yeah that makes it look a lot cleaner. Thanks.

sasdata/transforms/test_interpolation.py
rprospero: Technically, the comment you marked was removed, but it's was still in the original code.  I've added a flag to pytest `--show_plots` that will cause the diagnostic plots to be produced.

sasdata/transforms/test_interpolation.py
rprospero: I went ahead and moved the file into the main test directory.

jamescrake-merani (approved):
Thank you for your work. It all looks good to me.  I was having an issue showing the plots but this is likely an issue with my machine, and not this PR.

PR #110, opened 2025-03-17
Rework Metadata as Dataclasses

rprospero:
I've reworked the Metadata into a a series of dataclasses.  These dataclasses are inspired by the CANSas v1.1 standard as it is likely to be the most comprehensive of the various types of metadata that my be included.  Thus we can expect that the metadata from other formats should be a subset of the metadata for CANSas.

I've also included a unit test to check the metadata of several of the files.

### DIFF
--- a/sasdata/metadata.py
+++ b/sasdata/metadata.py
@@ -158,20 +158,16 @@ class Sample:
         # return _str
 
 
+@dataclass(kw_only=True)
 class Process:
     """
     Class that holds information about the processes
     performed on the data.
     """
-    def __init__(self, target_object: AccessorTarget):
-        self.name = StringAccessor(target_object, "name")
-        self.date = StringAccessor(target_object, "date")
-        self.description = StringAccessor(target_object, "description")
-
-        #TODO: It seems like these might be lists of strings, this should be checked
-
-        self.term = StringAccessor(target_object, "term")
-        self.notes = StringAccessor(target_object, "notes")
+    name : Optional[ str ]
+    date : Optional[ str ]
+    description : Optional[ str ]
+    term : Optional[ str ]
 
     def single_line_desc(self):
         """
@@ -181,11 +177,10 @@ class Process:
 
     def summary(self):
         return (f"Process:\n"
-                f"    Name: {self.name.value}\n"
-                f"    Date: {self.date.value}\n"
-                f"    Description: {self.description.value}\n"
-                f"    Term: {self.term.value}\n"
-                f"    Notes: {self.notes.value}\n"
+                f"    Name: {self.name}\n"
+                f"    Date: {self.date}\n"
+                f"    Description: {self.description}\n"
+                f"    Term: {self.term}\n"
                 )
 
 class TransmissionSpectrum:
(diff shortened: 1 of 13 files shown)


sasdata/metadata.py
```diff
@@ -1,289 +1,157 @@
...
-                                                    "wavelength_max.units",
-                                                    default_unit=units.angstroms)
-
-        # Wavelength spread [float] [Angstrom]
-        # Quantity because it might have other units, such as percent
-        self.wavelength_spread = QuantityAccessor[float](target_object,
-                                                         "wavelength_spread",
-                                                         "wavelength_spread.units",
-                                                         default_unit=units.angstroms)
+    radiation: str
```
jamescrake-merani, line 101: Why was this not made optional? Is it assumed that all files with metadata will have this specified?

sasdata/temp_hdf5_reader.py
```diff
@@ -93,33 +99,159 @@ def connected_data(node: SASDataGroup, name_prefix="") -> list[NamedQuantity]:
...
+    """Pull a single quantity with length units out of an HDF5 node"""
+    magnitude = node.astype(float)[0]
+    unit = node.attrs["units"]
+    return Quantity(magnitude, units.symbol_lookup[unit])
+
+def parse_string(node : HDF5Group) -> str:
+    """Access string data from a node"""
+    return node.asstr()[0]
+
+def opt_parse(node, key, subparser):
```
jamescrake-merani, line 127: Would it be possible to type hint these functions?

sasdata/temp_hdf5_reader.py
```diff
@@ -93,33 +99,159 @@ def connected_data(node: SASDataGroup, name_prefix="") -> list[NamedQuantity]:
...
+def parse_sample(node : HDF5Group) -> Sample:
+    name = attr_parse(node, "name", parse_string)
+    sample_id = opt_parse(node, "ID", parse_string)
+    thickness = opt_parse(node, "thickness", parse_quantity)
+    transmission = opt_parse(node, "transmission", lambda n: float(n[0].astype(str)))
+    temperature = opt_parse(node, "temperature", parse_quantity)
+    position = opt_parse(node, "position", parse_vec3)
+    orientation = opt_parse(node, "orientation", parse_rot3)
+    details : list[str] = [node[d].asstr()[0] for d in node if "details" in d]
+    return Sample(name=name, sample_id=sample_id, thickness=thickness, transmission=transmission, temperature=temperature, position=position, orientation=orientation, details=details)
```
jamescrake-merani, line 224: This line is quite long. Maybe the arguments could be broken down into multiple lines?

jamescrake-merani (changes requested):
Code mostly LGTM. Just a few comments to address.

sasdata/metadata.py
```diff
@@ -1,289 +1,157 @@
...
+    beam_shape: Optional[str]
```
jamescrake-merani, line 102: I just noticed after my review while doing some other work that my editor is claiming the `Optional` keyword is deprecated, and that you should use something like the following instead.

```str | None```

Though interestingly, this isn't what the Python documentation says. Instead, it just says that since Python 3.10, the two statements are equivalent.

Since we're using them quite heavily here, its probably something to think about. I think I do prefer the `Optional` keyword used here but if it is indeed deprecated then we probably shouldn't use it.

sasdata/metadata.py
rprospero: Good catch!

rprospero: Also made me realise that `source` in `Instrument` should also be optional

sasdata/temp_hdf5_reader.py
rprospero: Again, good call.  Doing the type hints made me catch an issues with how `attr_parse` was working.

sasdata/temp_hdf5_reader.py
rprospero: I've broken down multiple longer lines so they all fit in a single tab on my machine.

sasdata/metadata.py
rprospero: I went ahead and switched out `Optional`

jamescrake-merani (approved):
It doesn't look like you have pushed yet but based on your comments, it seems everything is resolved. Feel free to merge but make sure you have pushed first.

PR #111, opened 2025-03-19
Fullname unit parser fix

rprospero:
This fixes an issue where the unit parser could not handle full unit names.  I've added a unit test to catch regressions.


jamescrake-merani (approved):
LGTM

PR #112, opened 2025-04-08
Sasdata xml parser

rprospero:
This adds XML parsers for CanSAS data to the new SasData format.  I've included a test to check for regressions.  While the parser is specifically setup to continue on the the event that it encounters unknown units (with a suitable warning for the user to ensure that they are aware of the issue), I have also made some small changes to the units parser to enable it to handle "microns" and "micrometer", both of which occured during tests, felt faily common, and previously caused unit parsing to fail.

### DIFF
--- a/sasdata/temp_xml_reader.py
+++ b/sasdata/temp_xml_reader.py
@@ -268,11 +268,16 @@ def load_data(filename) -> dict[str, SasData]:
     for entry in tree.getroot().findall(f"{version}:SASentry", ns):
         name = attr_parse(entry, "name")
 
+        title = opt_parse(entry, "Title", version, parse_string)
+
         if name is None:
-            name = f"SasData{dataindex:02}"
+            if title is None:
+                name = f"SasData{dataindex:02}"
+            else:
+                name = title
 
         metadata = Metadata(
-            title=opt_parse(entry, "Title", version, parse_string),
+            title=title,
             run=all_parse(entry, "Run", version, parse_string),
             instrument=opt_parse(entry, "SASinstrument", version, parse_instrument),
             process=all_parse(entry, "SASprocess", version, parse_process),
--- a/sasdata/temp_xml_reader.py
+++ b/sasdata/temp_xml_reader.py
@@ -33,12 +33,12 @@ ns = {
 }
 
 
-def parse_string(node: etree.Element, _version: str) -> str:
+def parse_string(node: etree._Element, _version: str) -> str:
     """Access string data from a node"""
     return "".join(node.itertext())
 
 
-def parse_quantity(node: etree.Element, version: str) -> Quantity[float]:
+def parse_quantity(node: etree._Element, version: str) -> Quantity[float]:
     """Pull a single quantity with length units out of an XML node"""
     magnitude = float(parse_string(node, version))
     try:
@@ -51,7 +51,7 @@ def parse_quantity(node: etree.Element, version: str) -> Quantity[float]:
     return Quantity(magnitude, unit)
 
 
-def attr_parse(node: etree.Element, key: str) -> str | None:
+def attr_parse(node: etree._Element, key: str) -> str | None:
     """Parse an attribute if it is present"""
     if key in node.attrib:
         return node.attrib[key]
@@ -59,10 +59,10 @@ def attr_parse(node: etree.Element, key: str) -> str | None:
 
 
 def opt_parse[T](
-    node: etree.Element,
+    node: etree._Element,
     key: str,
     version: str,
-    subparser: Callable[[etree.Element, str], T],
+    subparser: Callable[[etree._Element, str], T],
 ) -> T | None:
     """Parse subnode if preset"""
     if (inner_node := node.find(f"{version}:{key}", ns)) is not None:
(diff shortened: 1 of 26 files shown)


sasdata/temp_xml_reader.py
```diff
@@ -0,0 +1,297 @@
...
+                    unit_parser.parse(value.attrib["unit"])
+                    if "unit" in value.attrib
+                    else unitless
+                )
+                us[name] = unit
+            struct[name] = float(parse_string(value, version))
+            keys.add(name)
+        aos.append(struct)
+
+    # Convert array of structures to strucgture of arrays
```
jamescrake-merani, line 211: I think there is a typo here.

sasdata/temp_xml_reader.py
```diff
@@ -0,0 +1,297 @@
...
+    if len(sys.argv) > 1:
+        files = sys.argv[1:]
+    else:
+        files = [test_file]
+
+    for f in files:
+        print(f)
+        data = load_data(f)
+
+        # for dataset in data.values():
```
jamescrake-merani, line 296: Does this need to stay commented out? As otherwise it seems the data will be loaded in but nothing will actually be printed about the data itself for testing (only the filename).

sasdata/temp_xml_reader.py
```diff
@@ -0,0 +1,297 @@
...
+
+def load_data(filename) -> dict[str, SasData]:
+    loaded_data: dict[str, SasData] = {}
+    tree = etree.parse(filename)
+    root = tree.getroot()
+
+    version: str | None = None
+
+    # Find out cansas version
+    for n, v in ns.items():
```
jamescrake-merani, line 244: Could the logic in this for loop be abstracted into a separate function? So something like `get_cansas_version`, or whatever name you think is appropriate.

sasdata/temp_xml_reader.py
```diff
@@ -0,0 +1,297 @@
...
+        result[k] = Quantity(np.array(soa[k]), us[k])
+        if k + "dev" in uncertainties:
+            result[k] = result[k].with_standard_error(
+                Quantity(np.array(soa[k + "dev"]), us[k + "dev"])
+            )
+
+    return result
+
+
+def load_data(filename) -> dict[str, SasData]:
```
jamescrake-merani, line 236: `filename` is missing a type hint. I assume its a string?

jamescrake-merani (changes requested):
Overall the code looks very clean to me. Just some minor suggestions.

sasdata/temp_xml_reader.py
```diff
@@ -0,0 +1,330 @@
...
+    """Parse a radiation source"""
+    radiation = opt_parse(node, "radiation", version, parse_string)
+    beam_shape = opt_parse(node, "beam_shape", version, parse_string)
+    beam_size = opt_parse(node, "beam_size", version, parse_beam_size)
+    wavelength = opt_parse(node, "wavelength", version, parse_quantity)
+    wavelength_min = opt_parse(node, "wavelength_min", version, parse_quantity)
+    wavelength_max = opt_parse(node, "wavelength_max", version, parse_quantity)
+    wavelength_spread = opt_parse(node, "wavelength_spread", version, parse_quantity)
+    return Source(
+        radiation=radiation,
```
jamescrake-merani, line 135: I think it would be more concise to do something like 
```python
return Source(
    radiation=opt_parse(...)
```

Rather than creating a separate variable, and then plugging it in. In my opinion this wouldn't impact the readability.

jamescrake-merani: Like you have done in `parse_detector`

sasdata/temp_xml_reader.py
```diff
@@ -0,0 +1,330 @@
...
+        pixel_size=opt_parse(node, "pixel_size", version, parse_vec3),
+        slit_length=opt_parse(node, "slit_length", version, parse_quantity),
+    )
+
+
+def parse_aperture(node: etree._Element, version: str) -> Aperture:
+    """Parse an aperture description"""
+    size = opt_parse(node, "size", version, parse_vec3)
+    if size and (innerSize := node.find(f"{version}:size", ns)) is not None:
+            size_name = attr_parse(innerSize, "name")
```
jamescrake-merani, line 162: The indentation is looking inconsistent here as its not in line with `size_name = None` on the else statement. Unless I'm missing something?

sasdata/temp_xml_reader.py
jamescrake-merani: The version I'm seeing is still missing a type hint.

jamescrake-merani (changes requested):
Some more suggestions for you to consider.

sasdata/temp_xml_reader.py
rprospero: Good call

sasdata/temp_xml_reader.py
rprospero: I have no idea how that happened, but I've fixed it.

jamescrake-merani (approved):
I'm approving this now. Feel free to merge if you're happy with everything.

PR #113, opened 2025-04-15
Access raw data in metadata

rprospero:
# This PR should not be merged until after #112

Given the large number of data sources out there which are not strictly in compliance with the specifications, Lucas had established a system where users could still access arbitrary data from the raw file.  Unfortunately, my refactoring of the metadata had eliminated this critical functionality, so this PR reinstates it.

Essentially, appended to the metadata is a tree shaped copy of the file.  This tree can then be searched for any arbitrary key, which might be returned by string or as a Quantity.  The user could then perform activities such as plotting fitting results as a function of these arbitrary values.


sasdata/metadata.py
```diff
@@ -171,6 +187,44 @@ def summary(self):
...
+            if type(self.contents) is str:
+                children = f"\n{header}  {self.contents}"
+            else:
+                children = "".join([n.to_string(header + "  ") for n in self.contents])
+        else:
+            children = ""
+
+        return f"\n{header}{self.name}:{attributes}{children}"
+    def filter(self, name: str) -> list[ndarray | Quantity | str]:
+        match self.contents:
```
jamescrake-merani, line 211: This match statement is quite repetitive, although I realise the alternative may be chaining lots of `isinstance` calls together. Still, I think that might be better than repeating this block:
```python
if name == self.name
    return [self.contents]
```

jamescrake-merani: Or if there's a way to have multiple conditions in a `case` statement that would be great but I'm not sure if this is possible in Python.

jamescrake-merani (changes requested):

sasdata/metadata.py
rprospero: It is possible to have multiple conditions on a single statement, but the syntax was *not* what I was expecting.  It uses a bitwise `|` instead of a normal python `or`.  I'm sure there's a deep parser reason for that, but it's still surprising.

jamescrake-merani: Yes that is quite surprising. Its good to know that it does exist in Python though.

jamescrake-merani (approved):
I'm approving. Feel free to merge once you've merged your other PR, and you're happy with everything here.

PR #114, opened 2025-04-29
Make wheels that contain everything SasView needs

llimeht:
As part of work to simplify the SasView build and distribution systems, we want to reach a stage where `pip install sasdata` is enough to bring everything that SasView needs to both build and run. That means the sasdata wheel needs to include the sphinx sources for its documentation and it may as well contain the user documentation too.

Getting output into wheels requires tweaks to the build-system and so this is a reasonable time for a bit of spring cleaning to address multiple issues:

 - moving to a declarative build system
 - dropping deprecated interfaces
 - publishing the wheels as artifacts for testing/inspection/usage


jamescrake-merani (approved):
LGTM

PR #116, opened 2025-05-07
New sesans reader

rprospero:
This PR loads SESANS files in the SES file format into SasView.  It also includes changes to the `summary` function of the Metadata class to print the aperture information.

The encoding information from the SESANS magnets is included as `Processing` metadata.  This was chosen both for its flexibility and because SES has gone through a dataprocessing pass to calculate the depolarization.

### DIFF
--- a/sasdata/metadata.py
+++ b/sasdata/metadata.py
@@ -63,10 +63,10 @@ class Aperture:
     type_: str | None
 
     def summary(self):
-        return (f"Aperture:\n"
-                f"  Name: {self.name}\n"
-                f"  Aperture size: {self.size}\n"
-                f"  Aperture distance: {self.distance}\n")
+        return (f"   Aperture:\n"
+                f"     Name: {self.name}\n"
+                f"     Aperture size: {self.size}\n"
+                f"     Aperture distance: {self.distance}\n")
 
 @dataclass(kw_only=True)
 class Collimation:
@@ -82,7 +82,8 @@ class Collimation:
         #TODO collimation stuff
         return (
             f"Collimation:\n"
-            f"   Length: {self.length}\n")
+            f"   Length: {self.length}\n"+
+            "".join([a.summary() for a in self.apertures]))
 
 @dataclass(kw_only=True)
 class BeamSize:
@@ -185,7 +186,7 @@ class Instrument:
         return (
             "\n".join([c.summary() for c in self.collimations]) +
             "".join([d.summary() for d in self.detector]) +
-            self.source.summary())
+            (self.source.summary() if self.source is not None else ""))
 
 @dataclass(kw_only=True)
 class MetaNode:
--- a/sasdata/metadata.py
+++ b/sasdata/metadata.py
@@ -79,7 +79,6 @@ class Collimation:
 
     def summary(self):
 
-        #TODO collimation stuff
         return (
             f"Collimation:\n"
             f"   Length: {self.length}\n"+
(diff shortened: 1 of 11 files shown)


sasdata/metadata.py
```diff
@@ -82,7 +82,8 @@ def summary(self):
         #TODO collimation stuff
```
jamescrake-merani, line 82: Does this todo comment still need to be here?

sasdata/temp_sesans_reader.py
```diff
@@ -0,0 +1,224 @@
...
+import numpy as np
+
+
+def parse_version(lines: list[str]) -> tuple[str, list[str]]:
+    header = lines[0]
+    m = re.search("FileFormatVersion\s+(\S+)", header)
+
+    if m is None:
+        raise FileContentsException(
+            "Alleged Sesans file does not contain File Format Version header"
```
jamescrake-merani, line 31: Is it normal to use the word 'alleged' in this context?

sasdata/temp_sesans_reader.py
```diff
@@ -0,0 +1,224 @@
...
+def parse_kvs_quantity(key: str, kvs: dict[str, str]) -> Quantity | None:
+    if key not in kvs or key + "_unit" not in kvs:
+        return None
+    return Quantity(value=float(kvs[key]), units=unit_parser.parse(kvs[key + "_unit"]))
+
+
+def parse_kvs_text(key: str, kvs: dict[str, str]) -> str | None:
+    if key not in kvs:
+        return None
+    return kvs[key]
```
jamescrake-merani, line 58: You could alternatively use the `get` method:

https://docs.python.org/3/library/stdtypes.html#dict.get

Using this would allow this function to just be reduced to one line.

```
return kvs.get(key)
```

sasdata/temp_sesans_reader.py
```diff
@@ -0,0 +1,224 @@
...
+            definition=None,
+            raw=parse_metanode(kvs),
+        ),
+        kvs,
+        parts[2],
+    )
+
+
+def parse_data(lines: list[str], kvs: dict[str, str]) -> dict[str, Quantity]:
+    from collections import defaultdict
```
jamescrake-merani, line 167: Why is this import deferred?

sasdata/temp_sesans_reader.py
```diff
@@ -0,0 +1,224 @@
...
+                units=unit,
+            )
+
+        data_contents[h] = Quantity(
+            value=np.asarray(points[h]),
+            units=unit,
+            standard_error=error,
+        )
+
+    if "SpinEchoLength" not in data_contents:
```
jamescrake-merani, line 198: These three if statements are a little bit repetitive. Perhaps consider something like:
```
for required in ["SpinEchoLength", "Depolarisation", "Wavelength"]:
    if required not in data_contents:
        raise FileContentsException(f"SES file missing {required}")
```

jamescrake-merani (changes requested):
Please see comments.

sasdata/metadata.py
rprospero: Nope.  That todo was todone.

sasdata/temp_sesans_reader.py
rprospero: I've don one better on this and removed the function entirely.  It was only in a couple of places, so I just used a call to `dict.get` there

sasdata/temp_sesans_reader.py
rprospero: No go reason.  I moved it to the top of the file.

sasdata/temp_sesans_reader.py
rprospero: You're absolutely right that that is the cleaner solution

jamescrake-merani (approved):
LGTM now :)

PR #120, opened 2025-06-04
Adding Alumina data for Size Distribution

butlerpd:
Added the data set being used for the unit tests and also used in the IRENA tutorials on how to use size distribution analysis. Eventually should be used for our own tutorial on the subject


krzywon (approved):
Looks reasonable

PR #121, opened 2025-06-17
Merge Sasdata 0.10.0 into master

krzywon:
This brings master up-to-date with the relase branch.


jamescrake-merani:
The CI failures are coming from a typo I made on the version tag. It seems hatch crashes when it gets a malformed version string. I thought I fixed this by deleting the old tag, but hatch is still trying to use it. I think the problem is that I deleted it locally, but that change doesn't apply to the remote repository. I'll try to fix this.

butlerpd (approved):

butlerpd:
approved as per discussion at meeting. Everything should have been reviewed for the release so this should be mostly a formality. We may in fact want to automate this in the future

PR #122, opened 2025-06-17
Quantity test fixes

rprospero:
This fixes the broken tests in the refactor branch as well as bringing in more tests

- A case type in temp_sesans_reader is fixed
- numerical_encode uses utf-8 strings for the value, instead of raw bytes.  This fixes a JSON serialisation bug (since bytes aren't serialised properly)
- A proper equality comparison method was included for the `Quantity` class.  This fixes `utest_sasdataload`.  Previously, the `Quantity` object default to reference equality, which cause the test to fail, despite the values being equal
- Added `latex_units_for_testing` which was lost during an earlier rebase
- Several test files were moved from the `sasdata/quantity` folder into the `test/quantities` folder.  The tests previous did not run during a call to `pytest`.  Adding in these tests caught the JSON serialisation bug mentioned above.
- Fixes an array dimension issue in the solver in meshmerge

### DIFF
--- a/sasdata/slicing/meshes/meshmerge.py
+++ b/sasdata/slicing/meshes/meshmerge.py
@@ -67,7 +67,18 @@ def meshmerge(mesh_a: Mesh, mesh_b: Mesh) -> tuple[Mesh, np.ndarray, np.ndarray]
 
     non_singular = np.linalg.det(deltas) != 0
 
-    st = np.linalg.solve(deltas[non_singular], start_point_diff[non_singular])
+    st = np.linalg.solve(
+        deltas[non_singular],
+        # Reshape is required because solve accepts matrices of shape
+        # (M) or (..., M, K) for the second parameter, but ours shape
+        # is (..., M).  We add an extra dimension to force our matrix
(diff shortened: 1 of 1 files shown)


sasdata/quantities/quantity.py
```diff
@@ -1187,6 +1187,12 @@ def in_si_with_standard_error(self):
         else:
             return self.in_si(), None
 
+    def __eq__(self: Self, other: Self) -> bool | np.ndarray:
+        if self.units == other.units:
+            return self.value == other.value
+        return self.value == other.in_units_of(self.units)
```
jamescrake-merani, line 1193: Could this method not be reduced to a one liner? Perhaps by using `or`. I think the following statement is logically equivalent:

```self.units == other.units or self.value == other.in_units_of(self.units)```

Although I could be wrong so please verify yourself.

jamescrake-merani (changes requested):
Mostly looks good. Just one comment.

jamescrake-merani (approved):
LGTM now. Feel free to merge if its ready.

PR #123, opened 2025-06-24
Added a function to load an ASCII file without specifying any parameters

jamescrake-merani:
This is designed to replicate the 'load data' button behaviour where the parameters, and dataset type are guessed from the file.

This branch also includes some ruff formatting, and cleanup.


sasdata/temp_ascii_reader.py
```diff
@@ -1,42 +1,59 @@
...
 from sasdata.quantities.quantity import Quantity
 from sasdata.quantities.accessors import Group
-from sasdata.data_backing import Dataset 
+from sasdata.data_backing import Dataset
 from enum import Enum
 from dataclasses import dataclass, field
 import numpy as np
 import re
 from os import path
+from os.path import basename
```
DrPaulSharp, line 24: Given the addition of this line, is the above line `from os import path` still required?

sasdata/temp_ascii_reader.py
```diff
@@ -1,42 +1,59 @@
...
 @dataclass
 class AsciiReaderParams:
     """This object contains the parameters that are used to load a series of
     ASCII files. These parameters can be generated by the ASCII Reader Dialog
     when using SasView."""
-    filenames: list[str] # These will be the FULL file path. Will need to convert to basenames for some functions.
+
+    filenames: list[
+        str
+    ]  # These will be the FULL file path. Will need to convert to basenames for some functions.
```
DrPaulSharp, line 47: I'm assume this change is due to the ruff formatter making sure the line is sufficiently short. Personally, I'd say that this particular change is undesirable, so I've got two things to consider. Firstly what value is used for line length for the ruff formatter, and secondly could the comment be either trimmed or moved to the line above?

DrPaulSharp (approved):
All looks good, I'd just like you to have a look at the line lengths please.

sasdata/temp_ascii_reader.py
jamescrake-merani: I realised that the `basename` call I added is inconsistent with the way that its called in the rest of the file (where `path.basename` is used) so I've fixed that.

sasdata/temp_ascii_reader.py
jamescrake-merani: I don't think we have a proper Ruff config yet so its just going off whatever Ruff's defaults are. I think the best solution to this for now is just to move the comment.

PR #125, opened 2025-07-07
Update the ASCII reader to import raw metadata

jamescrake-merani:
The ASCII reader was not changed to be made compatible with Adam's metadata refactor. This PR fixes that so that now ASCII metadata is loaded in as raw.

### DIFF
--- a/test/utest_temp_ascii_reader.py
+++ b/test/utest_temp_ascii_reader.py
@@ -140,9 +140,9 @@ def test_mumag_metadata():
                 assert datum.metadata.raw.filter("counting_index") == ["1"]
                 assert datum.metadata.raw.filter("applied_magnetic_field") == ["33"]
                 assert datum.metadata.raw.filter("saturation_magnetization") == ["1640"]
-                assert datum.metadata.raw.filter("demagnetizing_field") == ["22.874115"]
+                assert datum.metadata.raw.filter("demagnetizing_field") == ["22"]
             case "6_1270_1640_24.331065.csv":
                 assert datum.metadata.raw.filter("counting_index") == ["6"]
                 assert datum.metadata.raw.filter("applied_magnetic_field") == ["1270"]
                 assert datum.metadata.raw.filter("saturation_magnetization") == ["1640"]
-                assert datum.metadata.raw.filter("demagnetizing_field") == ["24.331065"]
+                assert datum.metadata.raw.filter("demagnetizing_field") == ["24"]


sasdata/temp_ascii_reader.py
```diff
@@ -194,9 +189,18 @@ def load_data(params: AsciiReaderParams) -> list[SasData]:
...
         )
+        metadata = Metadata(
+            title=None,
+            run=[],
+            definition=None,
+            sample=None,
+            instrument=None,
+            process=None,
+            raw=raw_metadata,
+        )
```
DrPaulSharp, line 203: Given the raw metadata, is it not possible to determine any of the other fields at this point?

test/utest_temp_ascii_reader.py
```diff
@@ -89,3 +99,50 @@ def test_ascii_2d():
...
+            case "1_33_1640_22.874115.csv":
+                assert datum.metadata.raw.filter("counting_index") == ["1"]
+                assert datum.metadata.raw.filter("applied_magnetic_field") == ["33"]
+                assert datum.metadata.raw.filter("saturation_magnetization") == ["1640"]
+                assert datum.metadata.raw.filter("demagnetizing_field") == ["22"]
+            case "6_1270_1640_24.331065.csv":
+                assert datum.metadata.raw.filter("counting_index") == ["6"]
+                assert datum.metadata.raw.filter("applied_magnetic_field") == ["1270"]
+                assert datum.metadata.raw.filter("saturation_magnetization") == ["1640"]
+                assert datum.metadata.raw.filter("demagnetizing_field") == ["24"]
```
DrPaulSharp, line 148: What about the test files 2-5?

test/utest_temp_ascii_reader.py
```diff
@@ -89,3 +99,50 @@ def test_ascii_2d():
...
+def test_mumag_metadata():
+    filenames = [
+        "1_33_1640_22.874115.csv",
+        "1_33_1640_22.874115.csv",
+        "2_42_1640_23.456895.csv",
+        "3_61_1640_23.748285.csv",
+        "4_103_1640_24.039675.csv",
+        "5_312_1640_24.331065.csv",
+        "6_1270_1640_24.331065.csv",
+    ]
```
DrPaulSharp, line 113: I'd suggest parameterising the test over the filenames, but I see the note at the top of this test module.

sasdata/temp_ascii_reader.py
```diff
@@ -147,22 +146,18 @@ def load_quantities(params: AsciiReaderParams, filename: str) -> dict[str, Quant
...
+        children = []
         for metadatum_name, metadatum in top_level_item.values.items():
-            children[metadatum_name] = Dataset(metadatum_name, metadatum, {})
-        # This is a special set which needs to live at the root of the group.
-        # TODO: the 'other' name will probably need to change.
+            children.append(MetaNode(name=metadatum_name, attrs={}, contents=metadatum))
         if top_level_key == "other":
-            root_children = root_children | children
+            root_contents.extend(children)
         else:
```
DrPaulSharp, line 157: Why does the "other" key need to be treated differently? Specifically, why do you add a list of MetaNodes in this case instead of a MetaNode with contents that is a list of MetaNodes for the remaining keys?

test/utest_temp_ascii_reader.py
jamescrake-merani: I did this for brevity, to stop the test from getting overly long, and repetitive.

test/utest_temp_ascii_reader.py
jamescrake-merani: I think that note may be a little outdated, as I made some changes to the ASCII reader to give it some more defaults, and reduce the amount of parameters that need to be passed to it. So this could

The reason why this particular test isn't parametrised is because these files need to be loaded together at the same time because they all have the same format for metadata, and I want to make sure the reader can handle this.

sasdata/temp_ascii_reader.py
jamescrake-merani: Good question. This is because there is an 'other' category of metadata in the reader GUI which I felt should sit at the root rather than being in a separate category.
image
But now I think about it, I don't think it makes sense to have this name, so I'm going to put it on my todo list to look into why it had that name in the first name, and if I can change it (possibly by having it at the root of the GUI in the screenshot as well, so its not in its own category).

test/utest_temp_ascii_reader.py
DrPaulSharp: Ok, but do they need to be included in the test at all in that case?

PR #135, opened 2025-07-18
Fix ASCII reader column guesses

jamescrake-merani:
This PR adds a fix to the ASCII tests so that they fail when no data is loaded. This is temporary solution until the unit test refactor has been completed.

There was an issue where guessed columns would have their unit set to the unit *group* rather than an *individual* unit. This PR introduces some rudimentary logic for guessing a default unit, but this can be expanded on in the future.

After fixing the ASCII test, a separate bug has been uncovered where 2D ASCII files can't be loaded because the reader can't currently guess where the data starts from properly. This has been marked as an expected fail, and will be fixed in a separate PR.

### DIFF
--- a/sasdata/default_units.py
+++ b/sasdata/default_units.py
@@ -5,11 +5,12 @@ import sasdata.quantities.units as unit
 from sasdata.dataset_types import unit_kinds
 
 default_units = {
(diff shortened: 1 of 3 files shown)


DrPaulSharp (approved):
All looks good.

PR #138, opened 2025-07-25
Add PyPi publishing step to workflow

krzywon:
This adds automated pypi publishing to the github workflow.

Currently untested, but I could create a temp tag to see if it works.


jamescrake-merani (commented):
The changes look reasonable to me, but its important we have this tested first. As such, I'm not approving yet.

i think creating a temp tag probably works.

PR #146, opened 2025-08-06
Applies fixes for isort (I) ruff ruleset

DrPaulSharp:


test/quantities/utest_operations.py
```diff
@@ -1,9 +1,19 @@
 import pytest
 
-from sasdata.quantities.quantity import Operation, \
-    Neg, Inv, \
-    Add, Sub, Mul, Div, Pow, \
-    Variable, Constant, AdditiveIdentity, MultiplicativeIdentity
```
jamescrake-merani, line 6: Again, I'm guessing this was done for readability, though I'm also not sure its worth bringing in an exception for. I'll leave that up to your judgement.

jamescrake-merani (approved):
Looks all good to me. Just one comment for you to think about.

PR #158, opened 2025-08-15
157 refactor design documentation

lucas-wilkins:
Start of documentation for design decisions


butlerpd:
> @lucas-wilkins I think this PR is pointing to the wrong branch It should be `refactor_24`?

Ahhh... that makes more sense. Also the title and explanation of what this PR is seems wrong? This is not about documenting design decisions as much as making the changes? A clearer statement of what the PR is doing would be helpful?

PR #160, opened 2025-08-18
PyPi Upload fix

krzywon:
The PyPi uploads were looking for the wrong file names. This should fix that issue, and extends the wheel creation beyond ubuntu.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,7 @@ jobs:
       fail-fast: false
 
     env:
-       BUILD_WHEEL: ${{ (matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12') && 'true' || '' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       fail-fast: false
 
     env:
-       BUILD_WHEEL: ${{ (matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12') && 'true' || '' }}
+       BUILD_WHEEL: ${{ matrix.python-version == '3.12' && 'true' || '' }}
```
jamescrake-merani, line 34: Out of interest, what does this `&& 'true'` do? Seems a bit odd, but I don't know much about GitHub CI, and I can see it was present in the original code.

jamescrake-merani (approved):

PR #147, opened 2025-08-07
Fixes whitespace errors

DrPaulSharp:

PR #166, opened 2025-10-23
Upgrade python version to 3.12 and fix linting errors

DrPaulSharp:
This PR upgrades the minimum version of python to 3.12 to prepare for merging in features of the refactoring project.

It also introduces the full ruff ruleset to the main branch, and fixes the linting errors in the code. These errors were fixed on the `refactor_24` branch in PRs: #140, #144, #146, #147, and the pre-commit hook was introduced in #164.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,6 @@
 # INSTALL.md recommends a venv that should not be committed
 venv
 .venv
-uv.lock
 
 # doc build
 /docs/sphinx-docs/build
--- a/uv.lock
+++ b/uv.lock
@@ -1,0 +1,1123 @@
+version = 1
+revision = 3
+requires-python = ">=3.12"
+
+[[package]]
+name = "alabaster"
+version = "1.0.0"
+source = { registry = "https://pypi.org/simple" }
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -25,6 +25,11 @@
 /dist
 .mplconfig
 
+# INSTALL.md recommends a venv that should not be committed
+venv
+.venv
+uv.lock
```
rprospero: Are we supposed to commit the lock file to keep versions the same for all developers?  I'm not that familiar with `uv`, so this is a genuine question.

rprospero (approved):
Looks good.  There was one thing that surprised me, but I'm marking this as approved because I don't know `uv` well enough to know if it's wrong.  Mostly, I'm used to other systems where the lock file is kept under version control so that all devs are sure to be on the exact same version.

Also interesting to see how many strings were still explicitly marked as unicode from the old Python 2 days.

DrPaulSharp:
I've merged SasView/sasview#3642, and updated to python 3.12 in SasView/sasview#3649

Issue #43, opened 2023-08-04
_Sector method can't average reliably for circular regions.


ehewins:
**Describe the bug**
The class `_Sector` in `sasdata/data_util/manipulations.py` runs into issues when called with values for the args `phi_min` and `phi_max` which are 2pi radians apart. This behaviour can be observed by following the steps below:

**To Reproduce**
1. In Sasview, Compute/Plot any 2D data from `example_data/2d_data`, using any model.
2. Right click the plot and select either 'Wedge Averaging in Q' or 'Wedge Averaging in Phi'.
3. Right click the plot again and select 'Edit Slicer Parameters', and set `delta_phi [deg]` to 180.
4. Try moving the central line around, and note the error in the terminal: `ValueError: Average Error: No points inside sector of ROI to average...`

Sometimes the bug doesn't occur immediately, but it will usually show itself after trying a few different `phi` values. If dragging the central line via the GUI, you'll notice that the slicer won't stay put when you release your mouse button.

**Additional context**
I believe the issue begins with lines 875 and 876 of `manipulations.py`. If `self.phi_min` and `self.phi_max` are 2pi radians apart, then after being passed through the `flip_phi()` function the new variables `phi_min` and `phi_max` have the same value. There are then a load of checks for each of the points in the dataset to see if they're in the region enclosed by `phi_min` and `phi_max`. Instead of examining a circular region for valid datapoints, the `_avg` method examines an infinitesimal region which finishes immediately after it starts. Hence the error message: `ValueError: Average Error: No points inside sector of ROI to average...`.

PR #174, opened 2025-11-14
Fix Sector average for circular regions

dehoni:
This is a targeted fix in sasdata/data_util/manipulations.py to handle the case for a full circle where phi_max is phi_min + 2*pi. Instead of letting flip_phi collapse those into the same value (which yields a zero-width angular interval), the new logic detects the original span that corresponds to the full circle and constructs a continuous interval for binning (phi_min .. phi_min + 2*pi). This preserves correct binning width for full-circle sectors.

Fixes https://github.com/SasView/sasdata/issues/43

### DIFF
--- a/sasdata/data_util/manipulations.py
+++ b/sasdata/data_util/manipulations.py
@@ -788,7 +788,7 @@ class Ring:
         :return: Data1D object
         """
         if data2D.__class__.__name__ not in ["Data2D", "plottable_2D"]:
-            raise RuntimeError("Ring averaging only take plottable_2D objects")        
+            raise RuntimeError("Ring averaging only take plottable_2D objects")
 
         # Get data
         data = data2D.data[np.isfinite(data2D.data)]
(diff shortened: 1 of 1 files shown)


sasdata/data_util/manipulations.py
```diff
@@ -937,7 +956,7 @@ def _agv(self, data2D, run='phi'):
             # Note that their values must not be altered, as they are used to
             # determine what points (also in the range 0, 2pi) are in the ROI.
             if phi_min > phi_max:
-                binning = Binning(phi_min, phi_max + 2 * np.pi, self.nbins, self.base)
+                binning = Binning(phi_min, phi_max + 2 * math.pi, self.nbins, self.base)
```
krzywon: Replace with `TwoPi` instead here and below? Would be easy if the definition was imported or defined outside a class/method/function.

dehoni: Replaced all occurences in the file with Pi and TwoPi.

sasdata/data_util/manipulations.py
```diff
@@ -920,8 +920,27 @@ def _agv(self, data2D, run='phi'):
         y_counts = np.zeros(self.nbins)  # Cycle counts (for the mean)
 
         # Get the min and max into the region: 0 <= phi < 2Pi
+        TwoPi = 2 * math.pi
```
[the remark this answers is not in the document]
dehoni: Done

krzywon:
Everything is resolved now in this PR so I will merge it, but this uses the quantities package which is part of the refactor. That makes including this into the sasview v6.1.2 release a bit tricky, so I will not make a separate sasdata release as part of the sasview release.

PR #178, opened 2025-11-17
Adding ND rebinner to refactor_24

pmneves7:
This pull request adds a new function called NDrebin to the transforms folder. The goal of NDrebin is to enable the rebinning of an arbitrary sampling of datapoints in N-dimensions to a regular grid in N-dimensions. This is highly useful for arbitrary data visualization when the data was not measured on a nice uniform set of points, as is the case for example with (qx,qy,qz) points measured on a SANS instrument.

Note that this function will somewhat average or interpolate data, so some resolution information is lost. The ideal case is to fit or model directly to measured datapoints, _then_ use NDrebin on **both** the data and the fit to produce plots and visualizations that can be compared.

Also added a utest_NDrebin file to test the NDrebin function. Additional manual tests can be performed by running the NDrebin.py file as __main__.

### DIFF
--- a/sasdata/transforms/NDrebin.py
+++ b/sasdata/transforms/NDrebin.py
@@ -190,7 +190,6 @@ class NDRebin:
 
     def __call__(self):
         self.run()
-        return self.binned_data
 
     def run(self) -> None:
         """Bin the data into the defined bins."""


sasdata/transforms/NDrebin.py
```diff
@@ -0,0 +1,687 @@
...
+        but instead contains the edges of the bins, so it is 1 longer
+        in each dimension
+    :step_size: is a list of Ndims numbers, contains the step size
+        along each dimensino
+    :num_bins: is a list of Ndims numbers, contains the number
+        of bins along each dimension
+
+
+    An example call might be:
+    # test syntax 1
```
krzywon: Adding `.. code-block::` will help with sphinx rendering of the example script.

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block

pmneves7:
Still refactoring the documentation and tests.

The new class-based syntax is:
```
rebin = NDRebin(I_ND, qmat,
    step_size=0.2*np.random.rand(Ndims)+0.1,
    lower=[1,2,3],
    upper=[9,8,7])
rebin.run()
Ibin = rebin.binned_data
qbin = rebin.bin_centers_list
```
        
I am wondering if the rebin.run() should just be automatically done when the class is constructed. To me this makes sense but maybe there are object oriented reasons not to?

pmneves7:
Should be ready to be reviewed again @krzywon

sasdata/transforms/NDrebin.py
```diff
@@ -0,0 +1,522 @@
...
+
+    def _calculate_bins(self):
+        # this is a non-vector way of binning the data:
+        # for ind in range(Nvals):
+        #     this_bin_ind = bin_inds[ind,:]
+        #     if not np.isnan(this_bin_ind).any():
+        #         this_bin_ind = this_bin_ind.astype(int)
+        #         binned_data[*this_bin_ind] = binned_data[*this_bin_ind] + data_flat[ind]
+        #         binned_data_errs[*this_bin_ind] = binned_data_errs[*this_bin_ind] + errors_flat[ind]**2
+        #         n_samples[*this_bin_ind] = n_samples[*this_bin_ind] + 1
```
DrPaulSharp: Is this section required, given the vector method below?

sasdata/transforms/NDrebin.py
```diff
@@ -0,0 +1,522 @@
...
+        self.bin_inds = None
+        self.binned_data = None
+        self.binned_data_errs = None
+        self.n_samples = None
+
+        self._prepared = False   # flag to avoid double-prepare
+
+    def __call__(self):
+        self.run()
+        return self.binned_data
```
DrPaulSharp: Why does the this routine return `self.binned_data`? Don't you want the functionality to be the same as `self.run`?

pmneves7:
> With respect to `rebin.run()`, the advantage of separating the run code from initialisation is that it enables the rebinner to be reused after it has been initialised. With further refactoring (which I am not suggesting for this PR) it would be possible to rebin with different parameters, for example:
> 
> ```
> rebin.run(nbins=30)
> rebin.run(nbins=40)
> ```
> 
> Given this, I'd suggest leaving the code as is regarding running the rebinner.
> 
> I've got a few comments, I'll let @krzywon do the full review.

This would still be the case if `rebin.run()` is called in `__init__` right? You could call `run` again later if you want to change the limits etc. Adding `run` to the initialization just saves one line when you define the `rebin` object the first time. In other words, when would someone want to create a rebin object and not want to then run it? Maybe I am overthinking this though, and if you think it is problematic we can keep it as is.

sasdata/transforms/NDrebin.py
```diff
@@ -0,0 +1,522 @@
...
+    binned_data_errs : 
+        has size num_bins and is NDimensional, contains
+        the propagated errors of the binned_data
+    bins_list : 
+        is a list of 1D vectors, is similar to bin_centers_list,
+        but instead contains the edges of the bins, so it is 1 longer
+        in each dimension
+    step_size : 
+        is a list of Ndims numbers, contains the step size
+        along each dimensino
```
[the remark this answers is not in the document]
pmneves7: fixed in latest commit

DrPaulSharp:
> > With respect to `rebin.run()`, the advantage of separating the run code from initialisation is that it enables the rebinner to be reused after it has been initialised. With further refactoring (which I am not suggesting for this PR) it would be possible to rebin with different parameters, for example:
> > ```
> > rebin.run(nbins=30)
> > rebin.run(nbins=40)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Given this, I'd suggest leaving the code as is regarding running the rebinner.
> > I've got a few comments, I'll let @krzywon do the full review.
> 
> This would still be the case if `rebin.run()` is called in `__init__` right? You could call `run` again later if you want to change the limits etc. Adding `run` to the initialization just saves one line when you define the `rebin` object the first time. In other words, when would someone want to create a rebin object and not want to then run it? Maybe I am overthinking this though, and if you think it is problematic we can keep it as is.

Hmm, given that you have written a `__call__` method, I think keeping the separation of functionality from initialising and running is the best approach here. Fundamentally the approach with the code as it stands allows for greater flexibility which is where the advantage is.

pmneves7:
> > @DrPaulSharp @krzywon Are we ready to merge this branch in? I think I addressed all of your concerns.
> 
> I'll take a proper look soon, but I need you to do a job for me first. Could you please rebase this branch on top of the refactor branch using `git rebase refactor_24` with the `refactor_24_NDrebin` branch checked out. I've been sorting the branches recently so this one has become out of date.

I think I have rebased the branch, though I am not a git ninja.

DrPaulSharp:
Ok, that sorts the rebase. I'll give the code a once over later today.

PR #180, opened 2025-11-17
Modifies autogen code to match ruff rules

DrPaulSharp:
The code used to autogenerate files in `_build_tables.py` has been modified to ensure the resulting files are ruff-compliant.


pmneves7 (commented):
I can confirm that running on my machine, this fixes the previous non-ruff compliant behavior. Running _build_tables now does not change anything unless you actually change the units.

PR #183, opened 2026-01-21
Full name for quantity

### LINKED ISSUE #97: Full name in quantities
Choose a good name to identify quantities so that variance calculations work sensibly. Probably include title, run and date in some way. So, the name used for unique identification of the Q quantity might be TITLE_DATE_RUN:Q

rprospero:
This adds a human readable unique id for the data sets in a dataset.  The general format that we use is "RUN TITLE:RUN NUMBERS:DATASET DESCRIPTION:BASE62 ENCODED HASH OF DATA".  As an example, the unique id for the Q data in `cansas1d.xml` is "Test title:1234:Q:440tNBqdx9jvci6CgjmrmD".

The hash is included to ensure that different data always have a different unique_id, even if they have identical metadata. Base 62 was chosen to provide a more compact representation (44% shorter) of the hash than just writing out the integer.

This closes #97.

### DIFF
--- a/sasdata/quantities/quantity.py
+++ b/sasdata/quantities/quantity.py
@@ -1114,7 +1114,6 @@ class Quantity[QuantityType]:
 
         self._id_header = id_header
         self.name = name
-        # print(f"ID Header: {self._id_header}, Quant: {self.value}")
 
     # TODO: Adding this method as a temporary measure but we need a single
     # method that does this.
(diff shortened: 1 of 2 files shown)


sasdata/quantities/quantity.py
```diff
@@ -1113,14 +1112,21 @@ def __init__(self,
 
         self.history = QuantityHistory.variable(self)
 
+        self._id_header = id_header
+        self.name = name
+        # print(f"ID Header: {self._id_header}, Quant: {self.value}")
```
DrPaulSharp: Is this line required?

Issue #172, opened 2025-11-14
Handling unrecognized units


pmneves7:
parse_unit (from sasdata/quantities/unit_parser.py) throws errors when given a unit it is not familiar with. This behavior can be reproduced by running parse_unit('foo') where the input string is any string that is not recognized.

A consistent policy should be developed for handling unfamiliar units.

My opinion: There should be some way of accommodating unknown units without crashing. This is important because, for example, metadata from different instruments may have either non-standard ways of representing units or non-standard units. It is important to include as much metadata as completely as possible under the FAIR data principle.

It is impractical to assume though that we can accommodate every single unit and all their conversions, and indeed sometimes it is necessary to work with a nonstandard unit.

My suggestion is that a non-standard unit should be treated as a new type of unit, entirely separate from the base SI classes. For example, if I have a pressure that is saved in metadata as "mm of mercury", if it is not included in _build_tables.py that 1 mmHg = 133.322387415 pascals, then treat "mm of mercury" as its own separate dimensional unit. It would not be equivalent to any other units. If you do math with it, you will get a derived quantity, so [1 mm of mercury] x [1 kg] = [1 mm of mercury x kg] and so on. Even [1 mm of mercury] x [1 pascal] = [1 mm of mercury / pascal]. This approach allows the unit to be included as metadata with a reported unit of some kind (so that it can be reported and plotted appropriately), and other values can be derived from it still if necessary.

jamescrake-merani:
@rprospero 's work linked to this issue has resolved it.

PR #190, opened 2026-03-03
172 unrecognized units (Again)

rprospero:
As mentioned in #172, users might want to track units that aren't necessarily SI measurements (e.g. Slices per Pizza, Pages per Book).  This PR adds an `ArbitraryUnits` class which can track these sorts of units while still interactive with the regular `Units and `NamedUnits` classes.

It also contains a modernisation of the `utest_units.py` test file so that the tests can be properly searched and controlled through pytest.

This is a continuation of #188.  A slip of the finger led to this branch being force pushed to `refactor_24`, which caused GitHub to believe that the PR was merged.  I fixed and rolled-back the branch, but it is a [known github issue](https://github.com/orgs/community/discussions/37189) that the PR cannot be restored after.

### DIFF
--- a/sasdata/quantities/_units_base.py
+++ b/sasdata/quantities/_units_base.py
@@ -420,7 +420,7 @@ class UnknownUnit(NamedUnit):
     def __rmul__(self: Self, other):
         return self * other
 
-    def __truediv__(self: Self, other: "Unit"):
+    def __truediv__(self: Self, other: "Unit") -> "UnknownUnit":
         match other:
             case UnknownUnit():
                 num = dict(self._numerator)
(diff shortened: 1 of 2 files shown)


sasdata/quantities/_units_base.py
```diff
@@ -308,57 +299,202 @@ def __eq__(self, other):
...
+                result = UnknownUnit(self._numerator, self._denominator)
+                result._unit /= other
+                return result
+            case _:
+                return NotImplemented
+
+    def __rtruediv__(self: Self, other: "Unit") -> "Unit":
+        return (self/other) ** -1
+
+    def __pow__(self, power: int | float) -> "Unit":
```
DrPaulSharp: Based on the code, the output here and in `__rtruediv__` should be `UnknownUnit`.

sasdata/quantities/_units_base.py
```diff
@@ -308,57 +299,202 @@ def __eq__(self, other):
...
+        num.sort()
+        den.sort()
+        return " ".join(num + den)
+
+    def __eq__(self, other):
+        match other:
+            case UnknownUnit():
+                return self._numerator == other._numerator and self._denominator == other._denominator and self._unit == other._unit
+            case Unit():
+                return not self._numerator and not self._denominator and self._unit == other
```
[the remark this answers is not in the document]
rprospero: Agreed

sasdata/quantities/_units_base.py
rprospero: Good catch

sasdata/quantities/units.py
krzywon, line 1: The more I look at `units.py`, the more I think this needs to be added to `.gitignore` and just let the build system worry about it. Why do we have the same content twice?

PR #196, opened 2026-03-19
Adds linear algebra operations

DrPaulSharp:
This PR adds linear algebra operations for quantities. Note that it introduces a new dependency, `sympy`, which has the BSD licence. https://www.sympy.org/en/index.html

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@ lxml
 # Calculation
 numpy
 scipy
+sympy
 
 # Unit testing
 pytest
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,6 @@ lxml
 # Calculation
 numpy
 scipy
-sympy
 
 # Unit testing
 pytest
--- a/sasdata/quantities/quantity.py
+++ b/sasdata/quantities/quantity.py
@@ -5,7 +5,6 @@ from typing import Any, Self, TypeVar, Union
 
 import h5py
 import numpy as np
-import sympy as sp
 from numpy._typing import ArrayLike
 
 from sasdata.quantities import units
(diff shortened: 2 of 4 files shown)


test/quantities/utest_operations.py
```diff
@@ -192,10 +254,31 @@ def test_matmul_evaluation(op, a, b, result):
...
+        (Norm_1, [[1, 2], [3, 4]], np.float64(6.0)),
+        (Norm_2, [[1, 2], [3, 4]], np.float64(np.sqrt(30.0))),
+        (Norm_2, [[1.0, 2.5], [3.0, 4.0]], np.float64(np.sqrt(32.25))),
+    ],
 )
-def test_transpose_evaluation(op, a, result):
+def test_axis_none_evaluation(op, a, result):
     f = op(Constant(a))
+    print(f.evaluate({}))
+    print(result)
```
krzywon: Please remove the print statements.

requirements.txt
```diff
@@ -5,6 +5,7 @@ lxml
 # Calculation
 numpy
 scipy
+sympy
```
krzywon: Adding another requirement is never ideal, especially for something we've never had to do before. If we haven't had to do this operation previously, is there a good justification for adding it now?

What are the real world use-cases for adding these operations?

sasdata/quantities/quantity.py
```diff
@@ -1044,6 +1151,72 @@ def __eq__(self, other):
...
+class Determinant(UnaryOperation):
+    """Array Determinant - backed by numpy's det method"""
+
+    serialisation_name = "determinant"
+
+    def evaluate(self, variables: dict[int, T]) -> T:
+        return np.linalg.det(self.a.evaluate(variables))
+
+    def _derivative(self, hash_value: int) -> Operation:
+        return Trace(sp.adjugate(self.a) * self.a._derivative(hash_value))
```
krzywon: Adding an entire dependency for a single use seems excessive. Are there other operation we might use the package for in the future?

krzywon (commented):
I'm conflicted on this one and will need to think more on it. Overall, it seems fine, but I think I need a real use case for having these operations with the data object. Is the plan to apply them to SAS data?

sasdata/quantities/quantity.py
DrPaulSharp: Hmm, I'm inclined to agree overall. It's better to use this dependency than write code for the adjugate operation but we only need it here and I've added the determinant as something that might be useful rather than having a definite use case. I can take the determinant operation out and put it on ice for now and we can put the rest in if that sounds like a good plan?

krzywon (approved):

PR #198, opened 2026-03-20
Refactor 24 trend

gonzalezma:
Proposition to modify the original Trend class. The idea is that one can define as many axes as desired, using a dict to store them. Each axis can be linked to a metadata (given a list with the path to the metadata as before) or generated "on the fly" (in this case one has also to give a list, with the constraints that the first element must NOT be a string and the number of elements in the list must be equal to the number of SasData objects in data. Are these limitations OK)?

If it is agreed that this is the right way, next steps should be to clean up a little bit the code and add new utests.


docs/source/user/data/data_formats_help.rst
```diff
@@ -55,7 +55,7 @@ separated by whitespaces or commas or semicolons, in the following order:
     
 where *Q* is assumed to have units of 1/Angstrom, *I(Q)* is assumed to have
 units of 1/cm, *dI(Q)* is the uncertainty on the intensity value (also as 1/cm),
-and *dQ(Q)* **is the one-sigma FWHM Gaussian instrumental resolution in** *Q*,
+and *dQ(Q)* **is the one-sigma Gaussian instrumental resolution in** *Q*,
```
krzywon, line 58: I don't think this change is correct. @butlerpd, @yunliu01, and/or @smk78, can you comment?

sasdata/metadata.py
```diff
@@ -458,14 +458,19 @@ def to_string(self, header=""):
             )
         else:
             attributes = ""
-        if self.contents:
-            if type(self.contents) is str:
+        match self.contents:
+            case str():
```
krzywon, line 462: The string, Quantity and ndarray cases are exactly the same as the default case. Maybe just remove all but the default?

sasdata/trend.py
```diff
@@ -1,48 +1,156 @@
...
+                    current_item = node
+                    break
         else:
-            return current_item.data
-    raise ValueError('End of path without finding a dataset.')
+            # Not a list, can't navigate further
+            raise ValueError('Path does not lead to a valid metadatum.')
+
+        if current_item is None:
+            raise ValueError('Path does not lead to a valid metadatum.')
```
krzywon, line 31: Duplicate errors. Combine?

sasdata/trend.py
```diff
@@ -1,48 +1,156 @@
...
+                continue
+
+            valid_data.append(datum)
+
+        # Update data with only valid items
+        self.data = valid_data
+
+        # Warn about filtered items
+        if invalid_indices:
+            print(f"Warning: Removed data items at indices {invalid_indices} - not SasData objects or missing/invalid metadata")
```
krzywon, line 83: The `print` statements should probably be changed to `logging.warn`.

sasdata/trend.py
```diff
@@ -1,48 +1,156 @@
...
+
+        if len(self.data) < 2:
+            print(f"Warning: Only {len(self.data)} valid data items remain")
+
+    def _validate_manual_values(self):
+        """Ensure manual value lists match data length"""
+        for axis_name, axis_config in self.trend_axes.items():
+            if isinstance(axis_config, list) and not isinstance(axis_config[0], str):
+                # This is a manual value list (not a path)
+                if len(axis_config) != len(self.data):
```
krzywon, line 97: Unnecessary nesting

sasdata/trend.py
```diff
@@ -1,48 +1,156 @@
...
+                        raise ValueError(f"trend_axes['{axis_name}'] path {axis_config} invalid for data item {i}: {e}")
+
+    def get_trend_values(self, axis_name: str) -> list:
+        """Get values for a named trend axis"""
+        if axis_name not in self.trend_axes:
+            raise KeyError(f"Axis '{axis_name}' not found")
+
+        axis_config = self.trend_axes[axis_name]
+
+        if isinstance(axis_config, list) and len(axis_config) > 0 and isinstance(axis_config[0], str):
```
krzywon, line 118: This check is already performed when creating the trend. Is this necessary?

docs/source/user/data/data_formats_help.rst
smk78: dQ(Q) values come from the Mildner & Carpenter equation. But what they actually represent depends if they come from reactor-SANS or a TOF-SANS; the former has 'Gaussian' wavelength distributions, the latter uses rectangular bins. @RichardHeenan discusses this in https://doi.org/10.1107/S0021889897002173, but see  https://docs.mantidproject.org/nightly/algorithms/TOFSANSResolutionByPixel-v1.html for how ISIS SANS and SNS SANS (and probably CSNS SANS) compute dQ(Q). So I would argue that neither the existing text, nor the proposed text, are sufficiently generic.
Maybe change to:
```
and *dQ(Q)* **is the one-sigma instrumental resolution in** *Q* and may be defined differently by different instruments (consult the instrument team),
```

gonzalezma: I agree with @smk78 correction. I removed the FWHM because the width of a Gaussian can be described using either sigma or the FWHM, but I think that one-sigma FWHM is just confusing.

sasdata/metadata.py
gonzalezma: Agreed.

sasdata/trend.py
gonzalezma: Agreed.

sasdata/trend.py
gonzalezma: Changed to raise explicit errors for each condition. To modify depending on what types of "manual axis" we want to accept.

sasdata/trend.py
gonzalezma: Agreed.

krzywon:
With the merge conflicts due to the rebase, I'm struggling to review this. I would suggest you merge the refactor branch into this one and fix and merge conflicts. Until then, I'm not sure what changes you made versus what came from the rebase.

gonzalezma:
Yes, I fully understand. I have created a new branch starting directly from the last refactor_24. The diffs for 2 files there (metadata.py and trend.py) are lost, but I think it is much clearer there and in any case I don't think that I would be able to recover this any better. Give a look to the new branch and then probably we can delete this one.

krzywon:
Closing in favor of #216

PR #201, opened 2026-04-13
199 remove autogen code

### LINKED ISSUE #199: Add `sasdata.quantities.units` to `.gitignore`
The unit conversion and unit handling system in the refactoring branch (`sasdata.quantities.units`) is a generated file, but also lives in the repo. When changes are made to the file that generates this file, but need to be committed, even though the changes in the base file nearly matches the changes in the generated file. Ideally, `units.py` should be added to `.gitginore` and removed from the repo entirely.

My concern with this is someone editing `units.py` instead of the base `_units.py` and then trying to generate the file which will destroy all of the changes.

rprospero:
Adding units.py to .gitignore can be a problem because the problem them becomes deciding *when* units.py gets generated.

Instead, I decided to punt on the issue by eliminating the autogenerated files entirely.  Instead, the code is dynamically added to the modules at runtime.  This means that the units module can never be out of sync and developers can't make changes that will be overwritten.

A couple of bits of advice for reviewing:
- It looks like I wrote a lot of code in __init__.py, but it's pretty much just the old code form _build_tables.py, with some minor chanes
  + Lines 167 and 463 changed from opening a file to putting everything in a StringIO buffer
  + In lines 171-172, I had to get the full path to the module code, since it's no longer running form code in the same directory
  + Lines 457-460 and 482-485 do the actual work of creating the modules
  + The use of `exec` is not a security concern here because it is not being called on untrusted strings.  Any attacker attempting to produce a security hole here would need to be able to modify __init__.py or _units_base.py, which would already provide them all the privileges that they could have gained from `exec`.
- The units.py and si.py files are still required to exist, since I need a module to add the dynamically generated functions onto.
- I removed accessors.py entirely, since the accessor system was deprecated
- I moved the unicode_superscript code into _units_base.py because that we the only place where the code was called and handling relative imports of dynamic imports was getting to be a pain.

The biggest disadvantage of this setup is that certain tools (e.g. MyPy) don't use the proper Python module loading system and do not see the members of the generated modules.  I also fully concede that there might be a cleaner architecture through the use of `getattr`, but that would be a significant rewrite and I thought that a smaller, incremental PR would be better at this time.

Closes #199


sasdata/quantities/__init__.py
```diff
@@ -0,0 +1,485 @@
+"""
+Builds a data file containing details of units
+"""
+
+import importlib
+import os.path
+import types
```
DrPaulSharp, line 7: This line should be removed as the module is unused.

DrPaulSharp (approved):
Thanks for the detailed review notes, I'm inclined to agree that this approach is an improvement over what we have and should be accepted on that basis, with us keeping an eye on possible improvements down the line.

krzywon (approved):
Seems reasonable. The `getattr` suggestion is probably better in the long-term, but this is a step in the right direction.

test/sasmanipulations/utest_averaging.py
```diff
@@ -119,6 +119,8 @@ def test_ring(self):
         """
             Test ring averaging
         """
+        print(self.data)
+        print(self.data.detector[0].beam_center)
```
krzywon, line 123: Did you mean to commit this?

DrPaulSharp: Argh, damnit. I'll throw a quick commit in.

PR #204, opened 2026-05-14
Remove exec magic from sasdata.quantities.units

rprospero:
My previous work on refactoring out the `units` module depended heavily on writing a module at run time and then running the code through `exec`.  While this removed the need to keep `_build_tables.py` and `units.py` in sync, it was also very ugly code that could be very frightening to edit.

This PR produces a modified version that edits an exiting `units` module instead of trying to write a new one from scratch.

* cleaner and easier to read than old code
  * No use of StringIO
  * No use of `exec`
  * Easier to follow data paths (e.g. the `fid.write(f"    '{name}': {name}")` pattern obscures the actual value in the dictionary).
* units module code isn't run if units module isn't called (the `exec` version would build the units module the moment someone imported `sasdata.quantities` instead of `sasdata.quantities.units`)

Again providing guidance:

* Everything before line 899 of `units.py` was just moved directly from `__init__.py`
* The variable name `this` was chosen to match convention from the Javascript community.  I had also considered using `self`, which is more Pythonic, but had the potential to be confusing, since it is usually attached to a class method.
* The remaining code is still a modification of `units.py`.  However, values are defined directly and moved into the module through `setattr`.


DrPaulSharp (approved):
This all looks good. I have one question, you noted that "Everything before line 899 of `units.py` was just moved directly from `__init__.py`". There is also code there that is copied from `_units_base.py` - did you intend to move that code, or is the copy correct?

rprospero:
@DrPaulSharp The comment was incorrect and the copy was intentional.  In Lucas' original setup, `_units_base.py` was read and then appended prepended to the final `units.py` file.  Similarly, in the *current* refactor_24, the file is again read in and prepended to the string passed to `exec`.  However, for this PR, since the beginning of `units.py` is just a normal module, I thought _units_base.py really no longer needed to be a separate file.

PR #156, opened 2025-08-14
Work on the SasData object definition

lucas-wilkins:
Reworked SasData classes

This introduces a new object, the Abscissa object, that describes how the axes data relates to ordinate data.

Issue #134, opened 2025-07-17
Rebinning and slicer support


lucas-wilkins:
Data operations corresponding to common slicer operations

PR #207, opened 2026-05-19
Refactor averaging.py and manipulations.py to use quantities

DrPaulSharp:
This PR updates the modules refactored in #47 to use `Quantity` rather than `Data1D`/`Data2D` objects. It should serve as an example of how we use these objects in the refactoring project.

### DIFF
--- a/sasdata/data_util/averaging.py
+++ b/sasdata/data_util/averaging.py
@@ -1,20 +1,19 @@
-"""
-This module contains various data processors used by Sasview's slicers.
-"""
+"""This module contains various data processors used by Sasview's slicers."""
 
 import math
 
 import numpy as np
 import numpy.typing as npt
 
+from sasdata.data import SasData
 from sasdata.data_util.binning import DirectionalAverage
 from sasdata.data_util.interval import IntervalType
 from sasdata.data_util.roi import CartesianROI, PolarROI
-from sasdata.dataloader.data_info import Data1D, Data2D
+from sasdata.dataloader.data_info import Data1D
 from sasdata.quantities.constants import Pi, TwoPi
 
 
-def get_dq_data(data2d: Data2D) -> npt.NDArray[np.floating]:
+def get_dq_data(data2d: SasData) -> npt.NDArray[np.floating]:
     """
     Get the dq for resolution averaging
     The pinholes and det. pix contribution present
@@ -23,23 +22,31 @@ def get_dq_data(data2d: Data2D) -> npt.NDArray[np.floating]:
     q = 0. Note This method works on only pinhole geometry.
     Extrapolate dqx(r) and dqy(phi) at q = 0, and take an average.
     """
-    z_max = max(data2d.q_data)
-    z_min = min(data2d.q_data)
-    dqx_at_z_max = data2d.dqx_data[np.argmax(data2d.q_data)]
-    dqx_at_z_min = data2d.dqx_data[np.argmin(data2d.q_data)]
-    dqy_at_z_max = data2d.dqy_data[np.argmax(data2d.q_data)]
-    dqy_at_z_min = data2d.dqy_data[np.argmin(data2d.q_data)]
+
+    q_data = np.sqrt(data2d._data_contents["Qx"].value**2 + data2d._data_contents["Qy"].value**2)
+
+    z_max = max(q_data)
+    z_min = min(q_data)
+
+    dqx_data = np.sqrt(data2d._data_contents["Qx"].variance.value)
+    dqy_data = np.sqrt(data2d._data_contents["Qy"].variance.value)
+
+    dqx_at_z_max = dqx_data[np.argmax(q_data)]
+    dqx_at_z_min = dqx_data[np.argmin(q_data)]
+    dqy_at_z_max = dqy_data[np.argmax(q_data)]
+    dqy_at_z_min = dqy_data[np.argmin(q_data)]
+
     # Find qdx at q = 0
     dq_overlap_x = (dqx_at_z_min * z_max - dqx_at_z_max * z_min) / (z_max - z_min)
     # when extrapolation goes wrong
-    if dq_overlap_x > min(data2d.dqx_data):
-        dq_overlap_x = min(data2d.dqx_data)
+    if dq_overlap_x > min(dqx_data):
+        dq_overlap_x = min(dqx_data)
     dq_overlap_x *= dq_overlap_x
     # Find qdx at q = 0
     dq_overlap_y = (dqy_at_z_min * z_max - dqy_at_z_max * z_min) / (z_max - z_min)
     # when extrapolation goes wrong
-    if dq_overlap_y > min(data2d.dqy_data):
-        dq_overlap_y = min(data2d.dqy_data)
+    if dq_overlap_y > min(dqy_data):
+        dq_overlap_y = min(dqy_data)
     # get dq at q=0.
     dq_overlap_y *= dq_overlap_y
 
@@ -47,8 +54,8 @@ def get_dq_data(data2d: Data2D) -> npt.NDArray[np.floating]:
     # Final protection of dq
     if dq_overlap < 0:
         dq_overlap = dqy_at_z_min
-    dqx_data = data2d.dqx_data[np.isfinite(data2d.data)]
-    dqy_data = data2d.dqy_data[np.isfinite(data2d.data)] - dq_overlap
+    dqx_data = dqx_data[np.isfinite(data2d._data_contents["I"].value)]
+    dqy_data = dqy_data[np.isfinite(data2d._data_contents["I"].value)] - dq_overlap
     # def; dqx_data = dq_r dqy_data = dq_phi
     # Convert dq 2D to 1D here
     dq_data = np.sqrt(dqx_data**2 + dqy_data**2)
@@ -56,9 +63,7 @@ def get_dq_data(data2d: Data2D) -> npt.NDArray[np.floating]:
 
 
 class Boxsum(CartesianROI):
-    """
-    Compute the sum of the intensity within a rectangular Region Of Interest.
-    """
+    """Compute the sum of the intensity within a rectangular Region Of Interest."""
 
     def __init__(self, qx_range: tuple[float, float] = (0.0, 0.0), qy_range: tuple[float, float] = (0.0, 0.0)) -> None:
         """
@@ -70,11 +75,11 @@ class Boxsum(CartesianROI):
         """
         super().__init__(qx_range=qx_range, qy_range=qy_range)
 
-    def __call__(self, data2d: Data2D) -> tuple[float, float, float]:
+    def __call__(self, data2d: SasData = None) -> tuple[float, float, float]:
         """
         Coordinate data processing operations and return the results.
 
-        :param data2d: The Data2D object for which the sum is calculated.
+        :param data2d: The SasData object for which the sum is calculated.
         """
         self.validate_and_assign_data(data2d)
         total_sum, error, count = self._sum()
@@ -97,7 +102,7 @@ class Boxsum(CartesianROI):
         data = weights * self.data
         # Not certain that the weights should be squared here, I'm just copying
         # how it was done in the old manipulations.py
-        err_squared = weights * weights * self.err_data * self.err_data
+        err_squared = (weights * self.err_data) ** 2
 
         total_sum = np.sum(data)
         total_errors_squared = np.sum(err_squared)
@@ -107,25 +112,22 @@ class Boxsum(CartesianROI):
 
 
 class Boxavg(Boxsum):
-    """
-    Compute the average intensity within a rectangular Region Of Interest.
-    """
+    """Compute the average intensity within a rectangular Region Of Interest."""
 
     def __init__(self, qx_range: tuple[float, float] = (0.0, 0.0), qy_range: tuple[float, float] = (0.0, 0.0)) -> None:
         """
         Set up the Region of Interest and its boundaries.
 
-        The units of these parameters are A^-1
         :param qx_range: Bounds of the ROI along the Q_x direction.
         :param qy_range: Bounds of the ROI along the Q_y direction.
         """
         super().__init__(qx_range=qx_range, qy_range=qy_range)
 
-    def __call__(self, data2d: Data2D) -> tuple[float, float]:
+    def __call__(self, data2d: SasData) -> tuple[float, float]:
         """
         Coordinate data processing operations and return the results.
 
-        :param data2d: The Data2D object for which the average is calculated.
+        :param data2d: The SasData object for which the average is calculated.
         """
         self.validate_and_assign_data(data2d)
         total_sum, error, count = super()._sum()
@@ -138,7 +140,7 @@ class SlabX(CartesianROI):
     Average I(Q_x, Q_y) along the y direction (within a ROI), giving I(Q_x).
 
     This class is initialised by specifying the boundaries of the ROI and is
-    called by supplying a Data2D object. It returns a Data1D object.
+    called by supplying a SasData object. It returns a Data1D object.
     The averaging process can also be thought of as projecting 2D -> 1D.
 
     There also exists the option to "fold" the ROI, where Q data on opposite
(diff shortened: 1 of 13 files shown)


sasdata/data_util/averaging.py
```diff
@@ -23,42 +24,48 @@ def get_dq_data(data2d: Data2D) -> npt.NDArray[np.floating]:
...
     Extrapolate dqx(r) and dqy(phi) at q = 0, and take an average.
     """
-    z_max = max(data2d.q_data)
-    z_min = min(data2d.q_data)
-    dqx_at_z_max = data2d.dqx_data[np.argmax(data2d.q_data)]
-    dqx_at_z_min = data2d.dqx_data[np.argmin(data2d.q_data)]
-    dqy_at_z_max = data2d.dqy_data[np.argmax(data2d.q_data)]
-    dqy_at_z_min = data2d.dqy_data[np.argmin(data2d.q_data)]
+
+    q_data = np.sqrt(data2d._data_contents["Qx"].value**2 + data2d._data_contents["Qy"].value**2)
```
krzywon, line 28: For `three_dim` data, would this slice across all Qz spaces?

sasdata/data_util/averaging.py
```diff
@@ -23,42 +24,48 @@ def get_dq_data(data2d: Data2D) -> npt.NDArray[np.floating]:
...
+
+    z_max = max(q_data)
+    z_min = min(q_data)
+
+    dqx_data = np.sqrt(data2d._data_contents["Qx"].variance.value)
+    dqy_data = np.sqrt(data2d._data_contents["Qy"].variance.value)
```
krzywon, line 34: Why the square root of the variance? Aren't the variances already the `dq` values?

sasdata/data_util/averaging.py
```diff
@@ -97,7 +104,7 @@ def _sum(self) -> tuple[float, float, float]:
         data = weights * self.data
         # Not certain that the weights should be squared here, I'm just copying
         # how it was done in the old manipulations.py
-        err_squared = weights * weights * self.err_data * self.err_data
+        err_squared = (weights * self.err_data) ** 2
```
krzywon: `weights * weights` is usually faster than `weights ** 2` for numpy arrays, which is why this was likely done this way originally.

sasdata/data_util/averaging.py
```diff
@@ -302,27 +312,27 @@ def __init__(
...
-        q_all = data2D.q_data[finite_mask]
-        qx_all = data2D.qx_data[finite_mask]
-        qy_all = data2D.qy_data[finite_mask]
-        err_all = data2D.err_data[finite_mask] if data2D.err_data is not None else None
-        mask_all = data2D.mask[finite_mask]
+        data_all = data2D.ordinate.value[finite_mask]
+        qx_all = data2D._data_contents["Qx"].value[finite_mask]
+        qy_all = data2D._data_contents["Qy"].value[finite_mask]
+        q_all = np.sqrt(data2D._data_contents["Qx"].value**2 + data2D._data_contents["Qy"].value**2)[finite_mask]
+        err_all = np.sqrt(data2D.ordinate.variance.value)[finite_mask]
```
krzywon: Again, why the square root?

sasdata/data_util/averaging.py
```diff
@@ -411,27 +422,32 @@ def __init__(
...
-        err_data = None
-        if data2D.err_data is not None:
-            err_data = data2D.err_data[np.isfinite(data2D.data)]
-        qx_data = data2D.qx_data[np.isfinite(data2D.data)]
-        qy_data = data2D.qy_data[np.isfinite(data2D.data)]
-        mask_data = data2D.mask[np.isfinite(data2D.data)]
+        valid_data = np.isfinite(data2D.ordinate.value)
+
+        data = data2D.ordinate.value[valid_data]
+        err_data = np.sqrt(data2D.ordinate.variance.value)[valid_data]
```
krzywon, line 449: One more square root of the variances. If these are all correct, feel free to resolve all of them, but I am noting each  I see to be sure all are corrected, if need be.

sasdata/data_util/manipulations.py
```diff
@@ -576,26 +536,28 @@ def _agv(self, data2D, run='phi'):
...
-        q_data = data2D.q_data[np.isfinite(data2D.data)]
-        err_data=None
-        if data2D.err_data is not None:
-            err_data = data2D.err_data[np.isfinite(data2D.data)]
-        qx_data = data2D.qx_data[np.isfinite(data2D.data)]
-        qy_data = data2D.qy_data[np.isfinite(data2D.data)]
-        mask_data = data2D.mask[np.isfinite(data2D.data)]
+        finite_mask = np.isfinite(data2D.ordinate.value)
+        data = data2D.ordinate.value[finite_mask]
+        err_data = np.sqrt(data2D.ordinate.variance.value)[finite_mask]
```
krzywon, line 551: sqrt.

sasdata/data_util/roi.py
```diff
@@ -26,106 +26,100 @@ def __init__(self, center: tuple[float, float] = (0.0, 0.0)):
...
+        else:
+            valid_data = np.isfinite(data2d.ordinate.value)
 
-        # Assign properties of the Data2D object to variables for reference
+        # Assign properties of the SasData object to variables for reference
         # during data processing.
-        self.data = data2d.data[valid_data]
-        self.err_data = data2d.err_data[valid_data]
+        self.data = data2d.ordinate.value[valid_data]
+        self.err_data = np.sqrt(data2d.ordinate.variance.value)[valid_data]
```
krzywon, line 59: sqrt

test/sasmanipulations/utest_averaging.py
```diff
@@ -37,25 +39,45 @@ def setUp(self):
...
-
-        source = data_info.Source()
-        source.wavelength = 10.0  # A
-        self.data.source = source
+        data_contents = {
+            "Qx": Quantity(np.arange(100), per_angstrom),
+            "Qy": Quantity(np.arange(100), per_angstrom),
+            "I": Quantity(x_0, per_centimeter),
+            "dI": Quantity(dx_0, per_centimeter)
+        }
```
krzywon, line 47: I'm curious to see what would happen with a `Qz` term in here. Will the calculation succeed?

krzywon (commented):
Most of my comments are related to three concepts.
- The differences between `three_dim` vs `two_dim` data and how the former would be handled in the current implementation. Ideally, the Qz values would be used if they exist.
- Variances and taking the square root of them versus using the bare values.
- x * x vs x ** 2

Overall, this is a good port, but bypasses the mesh paradigm added in the `sasdata.slicing` package. Where do you see that coming into play?

sasdata/data_util/averaging.py
DrPaulSharp: From line 1579 of `quantity.py` we have: `self._variance = standard_error**2`, where `standard_error` is the input of a `Quantity` object. I'll look through and see if there is the opportunity to refactor and use variances here if possible, but my understanding is that this is the correct transcription of the existing code.

krzywon: Okay, then you are doing the correct operation here, but from a fundamental standpoint, by taking the square of the standard error, aren't we changing the value of the data that should be immutable? And be squaring the data, how do we know which root is correct, the positive or the negative. Also, if future developers aren't aware of this, taking the raw variance will create an issue in our error propagation which will be difficult to debug.

Fundamentally, I'm not sure I agree with the underlying structure for many reasons, but this is something we can discuss.

DrPaulSharp: Yes, I'm having the same thoughts. I don't understand why the `Quantity` object takes in an error and records a variance. I'll ask around and see what I can find, and we can discuss.

DrPaulSharp: `Quantity` does have a `standard_deviation` method, which is:

```
def standard_deviation(self) -> "Quantity":
    return self.variance**0.5
```

which just seems to add to the mystery of why it's structured like this.

krzywon: In error propagation, the variances are used most often, so I think I understand the reasoning now, but we should be storing the standard_error as-presented, for reproducibility reasons.

sasdata/data_util/averaging.py
```diff
@@ -23,42 +24,48 @@ def get_dq_data(data2d: Data2D) -> npt.NDArray[np.floating]:
...
+    dqy_at_z_max = dqy_data[np.argmax(q_data)]
+    dqy_at_z_min = dqy_data[np.argmin(q_data)]
+
     # Find qdx at q = 0
     dq_overlap_x = (dqx_at_z_min * z_max - dqx_at_z_max * z_min) / (z_max - z_min)
     # when extrapolation goes wrong
-    if dq_overlap_x > min(data2d.dqx_data):
-        dq_overlap_x = min(data2d.dqx_data)
+    if dq_overlap_x > min(dqx_data):
+        dq_overlap_x = min(dqx_data)
```
pkienzle: You probably want np.min/np.max which operate directly on numpy arrays rather than builtin min/max which try the arrays as an iterable.

sasdata/data_util/averaging.py
```diff
@@ -23,42 +24,48 @@ def get_dq_data(data2d: Data2D) -> npt.NDArray[np.floating]:
...
     dq_overlap_y *= dq_overlap_y
 
     dq_overlap = np.sqrt((dq_overlap_x + dq_overlap_y) / 2.0)
     # Final protection of dq
     if dq_overlap < 0:
         dq_overlap = dqy_at_z_min
-    dqx_data = data2d.dqx_data[np.isfinite(data2d.data)]
-    dqy_data = data2d.dqy_data[np.isfinite(data2d.data)] - dq_overlap
+    dqx_data = dqx_data[np.isfinite(data2d.ordinate.value)]
+    dqy_data = dqy_data[np.isfinite(data2d.ordinate.value)] - dq_overlap
```
pkienzle: Save a little bit of time by computing the index once. Also, easier to maintain since you only have to update the index condition in one place.

sasdata/data_util/averaging.py
```diff
@@ -70,11 +77,11 @@ def __init__(self, qx_range: tuple[float, float] = (0.0, 0.0), qy_range: tuple[f
         """
         super().__init__(qx_range=qx_range, qy_range=qy_range)
 
-    def __call__(self, data2d: Data2D) -> tuple[float, float, float]:
+    def __call__(self, data2d: SasData = None) -> tuple[float, float, float]:
```
pkienzle: type is `SasData|None` here and elsewhere.

sasdata/data_util/averaging.py
```diff
@@ -302,27 +312,27 @@ def __init__(
...
 
-        data_all = data2D.data[finite_mask]
-        q_all = data2D.q_data[finite_mask]
-        qx_all = data2D.qx_data[finite_mask]
-        qy_all = data2D.qy_data[finite_mask]
-        err_all = data2D.err_data[finite_mask] if data2D.err_data is not None else None
-        mask_all = data2D.mask[finite_mask]
+        data_all = data2D.ordinate.value[finite_mask]
+        qx_all = data2D._data_contents["Qx"].value[finite_mask]
+        qy_all = data2D._data_contents["Qy"].value[finite_mask]
```
pkienzle: You shouldn't be peeking at the internals of SasData from outside the class. You could make `data_contents` a public attribute. Since users of the class need to compose `data_content` for the constructor this is appropriate. Or use `data2D.abscissae` to get an array of (qx, qy) points.

Note: `data2D.abscissae` makes a copy but `data2D.ordinate` does not. This will trip somebody up some day. Also, properties should not be doing a lot of work. Maybe turn `abscissae` into function call, or maybe cache it.

Your type system allows calling with 1D data. Are you sure you don't want a subclass of SasData that you know to be 2D? Then you can have qx,qy accessors and the (qx,qy) -> (q, phi) transforms built into the class.

For `q_all` below, you already have masked `qx_all` and `qy_all`. If you are not using abscissae with the 2-norm or a SasData2D subclass with a q method you can use qx_all and qy_all to calculate q.

Moving the masking operation into SasData seems appropriate. It is an operation that many users of the data will want. You want to be clear on the copy semantics: is it masking in place, or returning a masked copy, even if the copy contains everything.

Comments on the preexisting code:

It is suspicious that this slicer has an isfinite mask but others do not.

Why the `_all` tag on the variable names?

sasdata/data_util/averaging.py
```diff
@@ -334,24 +344,22 @@ def __call__(self, data2D: Data2D, ismask: bool = False) -> Data1D:
...
-            minor_axis=phi_axis,
-            lims=(major_lims, minor_lims),
-            nbins=self.nbins,
-            base=self.base,
-        )
+        directional_average = DirectionalAverage(major_axis=major_axis,
+                                                 minor_axis=phi_axis,
+                                                 lims=(major_lims, minor_lims),
+                                                 nbins=self.nbins,
+                                                 base=self.base)
```
pkienzle: I'm surprised by the indentation style. Doesn't ruff break after '(' and indent by 4?

sasdata/data_util/averaging.py
DrPaulSharp: Yes, I think that's right, though it would depend on how we set up the `abscissae` property for `qz`/refactor this code given @pkienzle suggestion below.

test/sasmanipulations/utest_averaging.py
DrPaulSharp: Just given it a go and it works. It works both if we keep it as 2D data that includes `qz`, and, with a bit of refactoring, for 3D data as well.

sasdata/data_util/averaging.py
DrPaulSharp: I think this was an IDEs call, but I've changed it to match the other calls to `DirectionalAverage`.

sasdata/data_util/averaging.py
DrPaulSharp: > You shouldn't be peeking at the internals of SasData from outside the class. You could make `data_contents` a public attribute. Since users of the class need to compose `data_content` for the constructor this is appropriate. Or use `data2D.abscissae` to get an array of (qx, qy) points.
> 
> Note: `data2D.abscissae` makes a copy but `data2D.ordinate` does not. This will trip somebody up some day. Also, properties should not be doing a lot of work. Maybe turn `abscissae` into function call, or maybe cache it.

This PR has shown that we need to refactor the `SasData` object. I am planning on doing that next in a separate PR.

> Your type system allows calling with 1D data. Are you sure you don't want a subclass of SasData that you know to be 2D? Then you can have qx,qy accessors and the (qx,qy) -> (q, phi) transforms built into the class.

Given the [SasData Design Decisions](https://github.com/SasView/sasdata/blob/refactor_24/sasdata/sasdata_design_decisions.rst), we do want the Data to be duck-typed. At present SasData has a `dataset_type` quantity that we can test to ascertain whether or not data is 2D. I'll keep this in mind going forward.

> For `q_all` below, you already have masked `qx_all` and `qy_all`. If you are not using abscissae with the 2-norm or a SasData2D subclass with a q method you can use qx_all and qy_all to calculate q.

Good plan, makes sense.

> Comments on the preexisting code:
> 
> It is suspicious that this slicer has an isfinite mask but others do not.
> 
> Why the `_all` tag on the variable names?

The other slicers get their `isfinite` masks through a call to the `validate_and_assign_data` routine in `roi.py`. This slicer is done in this way instead to preserve the `ismask` flag, which was the call made in #47. No idea on the `_all` tag. I'll remove it now, I wanted to keep things like this in place at first to make reviewing the ported code easier.

DrPaulSharp:
Thanks @krzywon and @pkienzle for your reviews. I propose merging this PR soon, and working on a new PR to refactor the `SasData` object based on the points raised here.

Good point about the `sasdata.slicing` package @krzywon. I'll need to investigate further how this should work with what we have here. The binning in `data_utils` that is used by the code here suggests we have a duplication of effort, but I'm not clear on how the new mesh types available can be incorporated with the slicers, or whether or not they need adding to - #134 leaves a bit to the imagination.

krzywon (approved):

krzywon:
We will talk about the different points on Friday during the technical meeting, if you are able to attend. We should merge this as-is and document the outstanding issues so they aren't lost.

butlerpd:
@DrPaulSharp is there a reason this has not yet been merged? it seems that was agreed a few weeks back?

DrPaulSharp:
@butlerpd I'll merge this soon. I've added another PR (#215) to go in before this one and I'm looking at the longstanding draft #156 with regard to @pkienzle comments.

sasdata/data_util/averaging.py
DrPaulSharp: There is work on an `Abscissa` class in #156. I will update that PR and work on these issues there.

Issue #211, opened 2026-06-03
build-time and development packages should not be in requirements.txt


llimeht:
`requirements.txt` is for specifying the runtime dependencies of the project - they are directly added to the dependencies of the wheel and so should not include packages that are for building the wheel or for developing the package.

build-time dependencies need to be in the `build-system.requires` list in `pyproject.toml`. Other packages that are useful for developing the module can go into a `requirements-dev.txt` or similar.

At present, the sasdata wheel accidentally (and incorrectly) depends on:

```
pytest
unittest-xml-reporting
sphinx
html5lib
pre-commit
```

It's not clear if `matplotlib` should really be a runtime requirement of `sasdata` although it is imported by `plotting`.

Marking this as blocking for the 0.12.x release as we do not want sasview to end up depending on `pre-commit`, `pytest` etc etc.

PR #212, opened 2026-06-03
Remove non-runtime dependencies from requirements.txt

llimeht:
Move development-time dependencies into `requirements-dev.txt`, to ensure that the `sasdata` wheel does not end up depending on testing tools.

Closes: #211


krzywon (approved):
This looks correct. I will let @jellybean2004, the release manager, decide whether or not this belongs in the next release.

jellybean2004:
As @llimeht has marked the issue as a blocker for the release, I will trust his judgment and add this to the next pre-release. If any major issues arise, I may have to omit it from the release.

PR #215, opened 2026-06-11
Refactors Quantity to have a standard error property

DrPaulSharp:
This PR refactors `quantity.py` to ensure Quantity objects have a `standard_error` property, with the `variance` property being simply `standard_error**2`. This makes it easier to access errors in other parts of the SasData code.

I propose merging this PR before #207, and then rebasing and rewriting that PR to incorporate the changes in this one.

### DIFF
--- a/sasdata/quantities/quantity.py
+++ b/sasdata/quantities/quantity.py
@@ -1531,9 +1531,9 @@ class QuantityHistory:
             operation(*[history.operation_tree for history in histories], **extra_parameters), references
         )
 
-    def has_variance(self):
+    def has_error(self):
         for key in self.references:
-            if self.references[key].has_variance:
+            if self.references[key].has_error:
                 return True
 
         return False
@@ -1570,19 +1570,15 @@ class Quantity[QuantityType]:
         self.hash_value = -1
         """ Hash based on value and uncertainty for data, -1 if it is a derived hash value """
 
-        self._variance = None
-        """ Contains the variance if it is data driven """
+        self._standard_error = standard_error
+        """ Contains the standard error if it is data driven """
 
         if standard_error is None:
             self.hash_value = hash_data_via_numpy(hash_seed, value)
         else:
-            self._variance = standard_error**2
             self.hash_value = hash_data_via_numpy(hash_seed, value, standard_error)
 
         self.history = QuantityHistory.variable(self)
-
-        self._id_header = id_header
-        self.name = name
 
         self._id_header = id_header
         self.name = name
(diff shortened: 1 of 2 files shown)


sasdata/quantities/quantity.py
```diff
@@ -1912,19 +1908,19 @@ def __init__(self, value: QuantityType, units: Unit, history: QuantityHistory):
...
+        return self._has_error
 
     @property
-    def variance(self) -> Quantity:
+    def standard_error(self) -> Quantity:
         if self._variance_cache is None:
             self._variance_cache = self.history.variance_propagate(self.units)
 
-        return self._variance_cache
+        return self._variance_cache**0.5
```
rprospero, line 1926: This function surprised me slightly and is making me think that I've misunderstood the PR.  I'd rather assumed that the `standard_error` property would just return the value of `_standard_error`, as defined in line 1573.

sasdata/quantities/quantity.py
```diff
@@ -1604,17 +1600,21 @@ def with_standard_error(self, standard_error: "Quantity"):
...
+        if self.has_error:
+            return Quantity(self._standard_error, self.units, name=self.name, id_header=self._id_header)
         else:
-            return Quantity(self._variance, self.units**2)
+            return Quantity(np.zeros_like(self.value), self.units, name=self.name, id_header=self._id_header)
 
+    @property
+    def variance(self) -> "Quantity":
+        """Get the variance of this object"""
+        return self.standard_error**2
```
rprospero, line 1617: It's a little odd that variance returns `standard_error`**2, but the `standard_error` property returns `self._variance_cache**0.5`.  Wouldn't it be better to cut out the middle man and just return `self._variance_cache` here?

rprospero (commented):
Just a bit of confusion on two points.

sasdata/quantities/quantity.py
DrPaulSharp: Yes, this is the case for the `standard_error` property of the `Quantity` object, defined on line 1607. Here is the `standard_error` property of the `DerivedQuantity` object. Note that the `__init__` for this class is:

```
class DerivedQuantity[QuantityType](Quantity[QuantityType]):
    def __init__(self, value: QuantityType, units: Unit, history: QuantityHistory):
        super().__init__(value, units, standard_error=None)

        self.history = history
        self._variance_cache = None
        self._has_error = history.has_error()
```

i.e., we call the `__init__` for a Quantity with `standard_error = None`. This is because for a `DerivedQuantity` we need to propagate errors from the quantity history, rather than read the property.

sasdata/quantities/quantity.py
DrPaulSharp: We can't do this here, because the standard error is defined differently for `Quantity` compared to `DerivedQuantity`, as noted above. We could write a `variance` property for the `DerivedQuantity` defined as you suggest, but I felt that the code duplication wasn't justified in this case despite the simpler operation. Happy to hear feedback on that point though.

rprospero (approved):

PR #216, opened 2026-06-11
Refactor 24 trend new

gonzalezma:
New trend branch created directly from refactor_24. It contains the changes done in March to the original Trend class to allow defining an arbitrary number of axes in the Trend (populated either from the metadata or by giving a list of numerical values) and new tests, as well as some improvements to address the review comments in the previous refactor_24_trend branch.


krzywon:
Yes, this is much easier to review. I'll close the other PR in favor of this one.

sasdata/metadata.py
```diff
@@ -458,14 +454,13 @@ def to_string(self, header=""):
             )
         else:
             attributes = ""
-        if self.contents:
-            if type(self.contents) is str:
-                children = f"\n{header}  {self.contents}"
-            else:
+        match self.contents:
```
krzywon, line 484: This match isn't completely equivalent to the previous if/else block. If `self.contents == ""`, the new check will give `children == f"\n{header}  {self.contents}"`, but gave an empty string before.

test/utest_trend.py
```diff
@@ -1,69 +1,151 @@
...
-    'FeNiB_perpendicular_Bersweiler_et_al',
-    'Nanoperm_perpendicular_Honecker_et_al',
-    'NdFeB_parallel_Bick_et_al'
+    "FeNiB_perpendicular_Bersweiler_et_al",
+    "Nanoperm_perpendicular_Honecker_et_al",
+    "NdFeB_parallel_Bick_et_al",
 ]
 
-custom_test_directory = 'custom_test'
+xml_file = path.join(path.dirname(__file__), "trend_test_data", "xml_test_files", "cansas1d_notitle.xml")
```
jamescrake-merani, line 20: Nowadays I prefer using importlib resources for this. Basically, it allows you to import the folder this test data folder as a regular Python module, and then read data from it.

See:
https://docs.python.org/3/library/importlib.resources.html#importlib.resources.path

But this is a very minor comment, as this implementation probably still works fine.

sasdata/trend.py
```diff
@@ -1,48 +1,182 @@
...
+        self._validate_metadata_paths()
+
+    def _filter_and_validate_data(self):
+        """Filter out non-SasData objects and validate data integrity"""
+        valid_data = []
+        invalid_indices = []
+
+        for i, datum in enumerate(self.data):
+            if not isinstance(datum, SasData):
+                invalid_indices.append(i)
```
jamescrake-merani, line 66: [ty](https://docs.astral.sh/ty/) (a Python type checker) is complaining that this code is unreachable, probably because `self.data` is type hinted to be a list of `SasData`, so therefore `datum` would never not be `SasData`. However, I appreciate that because Python is not strictly typed, its of course not impossible that `datum` is not `SasData`. But I think that if we get into type checking variables to make sure they're the type they should be, this could expand to cover all of the fields of these classes, at which point there are already libraries (e.g. [pydantic](https://pydantic.dev/docs/validation/latest/get-started/)) that handle this.

krzywon (approved):
This looks like it takes into multi-dimensional trends properly. Look at the one comment I made to be sure that doesn't affect functionality before merging.

sasdata/metadata.py
jamescrake-merani: I've just restored the old behaviour by adding a check for an empty string in the case. I'm not exactly sure which is the best, but that might be a decision to make later down the line.

jamescrake-merani (approved):
I added a few comments (forgot to submit the review so they didn't show up) but I think lets just get this merged in now, and we can deal with them later. They are very minor, and would probably require a bit of discussion.

jamescrake-merani:
I've just rebased to resolve the merge conflict. I'll check at the technical meeting, but if there are no objections then I will merge this PR.

PR #220, opened 2026-06-29
Add Abscissa class

DrPaulSharp:
This PR introduces the `Abscissa` class, which is able to determine the abscissa for given data types.


test/utest_sasmeasurement.py
```diff
@@ -0,0 +1,112 @@
...
+from sasdata.dataset_types import angle_dim, one_dim, three_dim, two_dim
+from sasdata.postprocess import deduce_qz
+from sasdata.quantities.constants import Pi
+from sasdata.quantities.quantity import Quantity
+from sasdata.quantities.units import none as unitless
+from sasdata.quantities.units import per_angstrom, per_centimeter, radians
+
+
+def test_1d(basic_metadata):
+    q = [1, 2, 3, 4, 5]
```
jamescrake-merani: For these tests: do you think it would be worth introducing some more test data?

